PR #70, opened 2014-12-03
add some keywords in cc-mode and text-mode

zhenglinj:
1.changed to google-c-style in cc-mode
2.added some keywords in text-mode to use in other mode


zhenglinj:
Yeah, you are right. I just want to make the keywords more easy to remember. I recommend that the style of c/c++ should be unified to google-c-style in cc-mode.

AndreaCrotti:
And another thing, I had some people complaining about normal words as snippets in text-mode so I would rather avoid them.

Could you please split the pull requests if you want to merge the C++ snippet changes?

zhenglinj:
#79

PR #75, opened 2015-01-20
Merge snippets from atotto for go mode

nanjj:
Copy benchmark, error, example, forrange, test, testmain from
https://github.com/atotto/yasnippet-golang. Changed test's key to `at`
to avoid conficts.


AndreaCrotti:
Don't use Go but looks good so I'll merge it in..

PR #76, opened 2015-01-28
dummy for not using `undefined`

s9gf4ult:


AndreaCrotti:
Are you sure the FIXME would have to be hard coded?
Is that what someone always writes in these cases?

Also to be picky would be great if you could rebase your branch before doing a PR, so that the Merge commits would not go in..
Thanks a lot anyway!

s9gf4ult:
Well, when you develop in top - down style you foten write some code like 

```haskell
func = do 
  stuff <- doStuff 
  if good stuff
    then return Foo
    else undefined
```

to implement that branch later, when you will know, that your types are correct and you must just fill dummies with code. Using `(error "FIXME: some error")` is more safe here, because when you see `undefined` error you dont know what to fix and where.

```haskell
func = do 
  stuff <- doStuff 
  if good stuff
    then return Foo
    else (error "FIXME: fix Bar branch")
```

This is usable just on development stage and not considered to be in production code.

PR #80, opened 2015-01-30
prog-mode/fixme: disable on sh-mode

stsquad:
It's cute the first time you try and close an if<->fi block in a shell
script but after the billionth time you just want it to stop doing that.


AndreaCrotti:
Good pont thanks, I think it's a more general issue, we don't need anything like  TODO/FIXME in all the inferior mode, because they are not buffers related to files anyway.

I'll see if there is a more general solution for that, thanks in the meanwhile

PR #81, opened 2015-03-13
added missing html snippets (recovered from yasnippet history)

julior:
The div tag was missing in the html snippets, and strangely there was a dov that looked like div. Probably a spelling issue.


julior:
Added another commit with extra html snippets recovered from yasnippet history...

AndreaCrotti:
So @julior the reason why they were removed is here
https://github.com/AndreaCrotti/yasnippet-snippets/pull/49

so in one way I agree they might be useful but on the other way emmet-mode is much better in this sense.
So I suggest you just try that instead, I would not recover these snippets ideally (but I can fix the typo of course).

julior:
Hi @AndreaCrotti  that makes sense. I didn't know about emmet-mode, which definitively looks much better. Probably, it would be good to mention about this in the Readme file.

Thanks

AndreaCrotti:
ok @julior I added more information in the README in this commit https://github.com/AndreaCrotti/yasnippet-snippets/commit/2baa8f24e7ea15cf5c0a418aeb493e432dd29f30
thanks

PR #83, opened 2015-04-23
Snippets for writing elisp tests.

raghavgautam:


emacs-lisp-mode/should
```diff
@@ -0,0 +1,6 @@
+# -*- mode: snippet -*-
+#contributor: Raghav Kumar Gautam
+#name: should
+#key: should
```
AndreaCrotti: About this, I'm not so sure it's so helpful to have a snippet which key is the same as the function itself.
Assuming you're using paredit/smartparens it would save no keystrokes at al, the deftest seems more useful but also that's quite long..

raghavgautam:
review comments have been addressed

AndreaCrotti:
Thanks

PR #87, opened 2015-05-09
Add some keywords in html-mode

ruajhonatan:


AndreaCrotti:
Thanks but have a look at
https://github.com/AndreaCrotti/yasnippet-snippets/blob/master/README.md#html-snippets
about HTML snippets and why they were removed..

You can add them to your own configuration just by placing in another directory and load snippets from both, how does that sound @ruajhonatan ?
thanks anyway

PR #89, opened 2015-05-18
Python and HTML additions

bgamari:
This adds a snippet to import `matplotlib.pyplot` and an HTML5 `DOCTYPE` annotation.


AndreaCrotti:
Merged thanks

PR #90, opened 2015-05-19
Comments for method should start with method name

jaigupta:
By go conventions, comments for functions should begin with the name of the function.


AndreaCrotti:
I don't really use go so I'm going to trust you on that, thanks

PR #92, opened 2015-05-26
Added doxygen-mode for cc-mode

henriquenj:
Resubmitting without the merge commits (thanks to @npostavs )

This pull request adds three new snippets for doxygen-like documentation following the manual available at http://www.stack.nl/~dimitri/doxygen/manual/docblocks.html. It also fix fopen snippet which was repeated on two modes.

Comments are welcome.

### DIFF
--- a/c-mode/fopen
+++ b/c-mode/fopen
@@ -1,4 +1,0 @@
-#name : FILE *fp = fopen(..., ...);
-# key: fopen
-# --
-FILE *${fp} = fopen(${"file"}, "${r}");
--- a/cc-mode/file_description
+++ b/cc-mode/file_description
@@ -1,0 +1,13 @@
+# -*- mode: snippet -*-
+#cotributor: Henrique Jung <henriquenj@gmail.com>
+#name: File description
+#key: \file
+#group: doxygen
+# --
+/**
+ *   \file ${1:`(file-name-nondirectory(buffer-file-name))`}
+ *   \brief ${2:A Documented file.}
+ ${3:*
+ *  ${4:Detailed description}
+ *
+}*/
--- a/cc-mode/function_description
+++ b/cc-mode/function_description
@@ -1,0 +1,14 @@
+# -*- mode: snippet -*-
+#cotributor: Henrique Jung <henriquenj@gmail.com>
+#name: Function description
+#key: \brief
+#group: doxygen
+# --
+/**
+ *  \brief ${1:function description}
+ ${2:*
+ *  ${3:Detailed description}
+ *
+ }*  \param ${4:param}
+ *  \return ${5:return type}
+ */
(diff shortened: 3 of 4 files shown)


c-mode/fopen
```diff
@@ -1,4 +0,0 @@
-#name : FILE *fp = fopen(..., ...);
```
AndreaCrotti: what was wrong with this?

cc-mode/.yas-parents
```diff
@@ -0,0 +1 @@
+doxygen-mode
```
AndreaCrotti: Here you're making a cycle of dependencies between cc-mode and doxygen-mode, is that desired?

is doxygen a minor mode anyway?
The dependency should only be one way imho..

AndreaCrotti:
Ok thanks @henriquenj I left a comment..

cc-mode/.yas-parents
henriquenj: You are correct, this is a mistake. I think my commit that remove it got lost in the merge.

Doxygen is a document generator that follows a particular syntax, a lot of popular free software projects use it. I put cc-mode as child of it because all C-like languages shares the same syntax. I don't think there's a doxygen minor mode for emacs though.

AndreaCrotti: @henriquenj I was checking ELPA and MELPA but there is no mention of doxygen-mode there..
I also used it in the past so I suppose it's maybe this http://doxymacs.sourceforge.net/ ?
The fact that it's not in the repo anywhere makes me suspect it's not really supported by anyone, is that possible?

henriquenj: As far as I know there's no doxygen mode for emacs. But the syntax is the same for all C-like languages (check the link I put on the pull request description), thats why I put cc-mode as parent. Maybe I should put inside cc-mode then?

AndreaCrotti: Ahh ok now it's clear sorry.
So well the question is @henriquenj, would you use doxygen with otoher modes that are not cc-mode (or inherited from it)?
If not just put the snippets there and set as "group: doxygen" so they are grouped..

Also because I think all the directories in this repository actually reflect a real major/minor mode, this would be an exception if we just add another directory for a mode that doesn't exist..

hope it's clear, thanks

henriquenj: I don't plan to use with other modes. But there are some corner cases, which is java-mode. Doxygen documentation specifies that Java also follows the C-like syntax, and java-mode is not a child of cc-mode. I don't use Java myself so I didn't include it.

For now, I think I'll submit all of them grouped inside cc-mode as you suggested. If someone want to move it to a separated mode, that can be made as well. I don't think doxygen use in Java is as widespread as it is in C/C++ though.

AndreaCrotti: Ok sounds coold @henriquenj, if you just move them in the cc-mode directory I'll merge this.. thanks

henriquenj: Done!

But I kinda sad there's no doxygen mode, maybe I will create one when I got the change.

AndreaCrotti:
Ok great thanks

PR #93, opened 2015-06-03
F90 rm stale typing snippets

zbeekman:
I won’t be offended if you don’t accept this PR, but these snippets are all super crusty or redundant. I’ve separated the commits into logical units and included the rational for the changes in the commit message.

If you don’t want to get rid of all of them, feel free to cherry pick commits. Also, if you like I could squash these all into one commit.

I’m planning to make some more updates to f90-mode snippets, but wanted to clean up some of the crusty-ness first.


AndreaCrotti:
@zbeekman well I don't really know much about Fortran so can't really judge, but why do you say they are crusty?
Are they just useless and noone would use them or what?

I'm happy to get rid of them if they are, but they were added by some other guy so would not want to break someone's set up either..

zbeekman:
@AndreaCrotti have you read through each of my commit messages? I think I do a decent job explaining the rational for removal there, but would be happy to elaborate further upon request.

The first four commits remove snippets for constructs that are both obsolete, or never officially part of the language AND poor programming practice. See my commit logs for details. Everything removed in the first four commits, would also likely be better expressed using a different Emacs paradigm like `abbrev-mode` `auto-completion` `hippie expand` or `dynamic abbreviation` since they consist only of one or two key words followed by `$0`. This seems like a waste or abuse of yasnippets. Other snippets suffer from this last defect too, but I am planning to modernize them and add more content to them. If you like we can leave these be, but unless a programmer has the distinct displeasure of working with 1960s and 1970s FORTRAN on 1960s and 1970s hardware I see little point in keeping these.

The last commit removes what are essentially duplicate snippets under a different `# key:` Code duplication is bad because then multiple copies of the same code must be maintained.

zbeekman:
@spacecowboy feel free to weigh in here, btw…

spacecowboy:
I'm only an infrequent Fortran programmer, but I agree with @zbeekman that most of what is removed is really not done anymore in modern Fortran.

Regarding the snippets that are copies under a different key, because I am coding infrequently in Fortran I can never remember abbreviations. Hence, I personally prefer the snippet to be under `program` and not `pr` for example. And since you kept *the right ones* I'm all for the changes :+1:

AndreaCrotti:
Yes ok sounds good to me, thanks!

PR #96, opened 2015-07-05
Add `mod` snippet to ruby-mode for `module` definitions

sshaw:
This is a patch from @jimeh [submitted to the yasnippet repo](https://github.com/capitaomorte/yasnippet/pull/282/) several years ago. It's missing from the current set of ruby snippets.


AndreaCrotti:
ok cool thanks!

PR #99, opened 2015-08-08
modernise the scala snippets, close #98

### LINKED ISSUE #98: scala snippets out of date
many of the scala snippets are only relevant for versions of scala that are no longer in common use (i.e. prior to 2.10, which itself is deprecated).

This is more of a note to self to go through and remove all the snippets that are no longer relevant.

fommil:
In https://github.com/ensime/ensime-emacs/ we have dramatically improved the experience of developing Scala code in Emacs. But many of the snippets in this package are interfering with functionality and, as a result, users (myself included) have preferred to turn off yasnippet in scala-mode.

This PR performs an audit of the snippets and has identified several classes:

1. deprecated features (e.g. scala actors, methods without explicit returns) --- removed these
2. encouraging bad code (e.g. runtime instanceof / casting, more relevant for older Scala) --- removed these
3. shortcuts for words that save a few characters --- removed these, as the functionality is much better served by https://github.com/company-mode/company-mode/issues/367 (these are probably the worst offenders)
4. relies on features from an ancient version of `scala-mode` (everybody uses `scala-mode2` now which is much better). This was mostly for scaladoc functionality... if anybody wants this again, please raise a ticket in `ensime-emacs` to make elisp functions available that can be used as drop-in replacements for the old `scala-mode` functions that no longer exist.

I also tweaked/rewrote a few of the existing snippets to match modern Scala style conventions and syntax.

The result is a small set of useful snippets that are relevant for modern scala!

We actually have some functionality in `ensime-mode` to generate test classes and so on, I'm wondering if perhaps we should define all that as snippets (CC @edani for thoughts). In which case, expect a PR with those templates added.

I have a snippet that I run when I create a new scala file, but it relies on an elisp function being present (to extract the username or copyright owner). I'm wondering if that is something that could be contributed also:

* https://github.com/fommil/dotfiles/blob/master/.emacs.d/init.el
* https://github.com/fommil/dotfiles/tree/master/.emacs.d/snippets/scala-mode


AndreaCrotti:
Oh thanks @fommil that's some awesome work..

About your last question, depending on some elisp function is not really an issue, just look at how it's done for Python here:
https://github.com/AndreaCrotti/yasnippet-snippets/blob/master/python-mode/.yas-setup.el

I'll review the changes and merge them soon, I just hope removing many snippets won't upset anyone, but as long as it's clear what's the alternative and why it was done I think we are fine..

fommil:
thanks! The guy who submitted a lot of these changes in the first place hasn't used emacs in years, so I doubt he'll complain.

fommil:
oh cool! I might package up my "template file" work at some point then and submit it to you. I can use a `dir-locals.el` to define the `copyright-owner` and it'll be used instead of the user's name (i.e. for work!)

fommil:
actually, I might also delete the `map`, `flatMap` and `foreach` snippets. I've just realised ENSIME does all this (with inferred types, so much better) and they do tend to conflict.

AndreaCrotti:
I guess the only question about that @fommil would be if there are people that use Emacs + Scala but not Ensime, that would still like to have these snippets..

Anyway I'm going to merge this since I suspect that this subset would be very small anyway, if anyone disagree please open another issue if you don't agree and we can find a solution..

AndreaCrotti:
Actually a possible solution would be to have these snippets enabled only if ensime-mode is not enabled.

However I'm not really sure how to do that, the other way around would be much easier (having more snippets when ensime-mode is enabled).

fommil:
@AndreaCrotti cool thanks. It's trivial to be able to add user-specific snippets (but a lot harder to exclude them) so I would advocate distributing the minimal set that is useful to everybody and then recommending user customisation beyond that.

AndreaCrotti:
Ok then it's merged for now, if I find another way to add them and making them available when ensime-mode is disabled even better..

fommil:
:smile: even if you do find a way to do that, I think a lot of the snippets are irrelevant in the light of company-mode. Yasnippet really shines when it is expanding templates, not just words.

PR #100, opened 2015-08-22
Installation instruction type in README

lmouelle:
The README requests that you use the non existent yas-root-directory, instead of the actual yas/root-directory. Changed character to / from - to save others an hour of head scratching.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ Otherwise if you want the latest and greatest snippets collection proceed as fol
 
 1. clone this repository
 2. add to your .emacs the following
-   - `(add-to-list 'yas-root-directory "$DIRECTORY_WHERE_YOU_CLONED")`
+   - `(add-to-list 'yas/root-directory "$DIRECTORY_WHERE_YOU_CLONED")`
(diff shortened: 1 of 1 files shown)


npostavs:
Rather talk about `yas-snippet-dirs`:

https://github.com/capitaomorte/yasnippet/blob/2f6cfef03139dbb74b0f4aeaffef0f9647bee5b8/yasnippet.el#L38-L49
```
;;       `yas-snippet-dirs'
;;
;;           The directory where user-created snippets are to be
;;           stored.  Can also be a list of directories.  In that case,
;;           when used for bulk (re)loading of snippets (at startup or
;;           via `yas-reload-all'), directories appearing earlier in
;;           the list override other dir's snippets.  Also, the first
;;           directory is taken as the default for storing the user's
;;           new snippets.
;;
;;           The deprecated `yas/root-directory' aliases this variable
;;           for backward-compatibility.
```

AndreaCrotti:
Yes better to talk about the correct way of doing things, thanks @npostavs and @lmouelle

PR #102, opened 2015-09-08
Add map and reduce to ruby-mode

nofxx:
Was missing those.


AndreaCrotti:
Ok thanks a lot @nofxx

PR #103, opened 2015-09-25
Add `clean` snippet to makefile-gmake-mode

gbalats:
Adding some snippets for `makefile-gmake-mode`:
* common functions (mostly about filesystem paths)
* a typical clean target
* a template definition (and generation).

Since the typical case (and the one worth to simplify by using yasnippet) is to have variable operands (e.g., `$(var)`), I used nested placeholder fields extensively to cover all cases (so that you only have to type `var` or skip the variable altogether and type in a constant instead).

### DIFF
--- a/makefile-gmake-mode/abspath
+++ b/makefile-gmake-mode/abspath
@@ -1,0 +1,6 @@
+# -*- mode: snippet; require-final-newline: nil -*-
+# name: abspath
+# contributor: gbalats
+# key: abs
+# --
+\$(abspath ${1:\$(${2:paths})})$0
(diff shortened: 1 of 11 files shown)


AndreaCrotti:
Thanks a lot @gbalats 
One question, are all these snippets only valid for Gnu Make or they are more generic?

Otherwise they can be added in makefile-mode if they are valid for any makefile..

gbalats:
I guess the snippet about a clean target is generic enough. The other ones rely on GNU extensions. I just noticed that `makefile-mode` is a parent mode of `makefile-gmake-mode`, so I will move just this snippet to the former.

gbalats:
I added a couple more snippets, neither of whom is generic enough for `makefile-mode`. Please review.

AndreaCrotti:
I'm not writing so many makefiles so haven't used all these features but I gave a look and they seem good to me, thanks!

PR #104, opened 2015-10-16
Generate idiomatic anonymous functions

### LINKED ISSUE #101: js-mode function snippet does not generate canonical javascript function syntax for anonymous functions
The javascript function snippet introduced here https://github.com/AndreaCrotti/yasnippet-snippets/commit/99e72dd2c291b92381c3750244fb90eae0ac5b18

When the `name` snippet parameter is skipped, the `function.el` snippet will generate a function expression like so

```js
function (one, two, three) {
//function body
}
```

but by far the most common style is to not have a space in between the function keyword and parameter list for anonymous functions:

```js
function(one, two, three) {
}
```

Almost any example you'll find on the web uses this style found here at mozdev on the function statement documentation:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/function

cowboyd:
This closes #101 per the discussion there.

### DIFF
--- a/js-mode/function
+++ b/js-mode/function
@@ -2,6 +2,6 @@
 # name: function
 # key: f
 # --
-function ${1:name}(${2:arg}) {
+function${1: ${2:name}}(${3:arg}) {
          $0
 }


AndreaCrotti:
Ok cool let's do this

PR #105, opened 2015-10-30
Add js2-mode snippets for jsdoc comments.

joeheyming:


js-mode/add-to-multiline
```diff
@@ -0,0 +1,7 @@
+# -*- mode: snippet -*-
+#name: multiline-comment
+#key: *
```
AndreaCrotti: I'm not entirely sure if we should add snippets that expand with so very simple keys.
* is a very used symbol, I probably would not make it a key for a snippet..

joeheyming: I can make it so we only can run this snippet when you are in a comment.
I find it very useful to add more lines to a comment.  Try it out and let me know if you don't like it.
Here is an example: http://imgur.com/xB9VIYa

npostavs: I usually use `M-j` (`indent-new-comment-line`) for this.

joeheyming: fwiw, I wanted the snippet key to be "* "  <- note the space.  But I couldn't figure out how to get that to work

joeheyming: @npostavs that is cool, I didn't know about that one

npostavs: In https://github.com/capitaomorte/yasnippet/pull/629 we had all agreed that nobody would ever want trailing whitespace in their snippet keys, so this is no longer possible (though I think you could still do it by calling `yas-define-snippet` directly)

joeheyming: Ok, I'm willing to drop the * rule since its too simple.  But I'm open for suggestions on how to get it to work properly (no hacks)

AndreaCrotti:
Ok I think they are specific enough not to conflict with anything else hopefully.. thanks

PR #112, opened 2016-01-15
added include-guard for c++-mode

hanslovsky:
I didn't find any snippet for include guards in c++-mode
For a header file named `<some-path>/include/package/header.extension`, this will expand `ig` to
```
#ifndef PACKAGE_HEADER_EXTENSION
#define PACKAGE_HEADER_EXTENSION

$0

#endif /* PACKAGE_HEADER_EXTENSION */
```

with `$0` being the cursor position.

### DIFF
--- a/c++-mode/include-guard
+++ b/c++-mode/include-guard
@@ -1,0 +1,10 @@
+# -*- mode: snippet -*-
+# name: include-guard
+# key: ig
+# --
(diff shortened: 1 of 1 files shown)


henriquenj:
@hanslovsky You can use the "once" snippet

hanslovsky:
Didn't check the cc-mode which contains the "once" snippet. This can be closed then.

PR #113, opened 2016-01-16
Move Ruby mod snippet to ruby-mode

sshaw:
21d5f45855f9683572cb519b48e344523b556271 added the `mod` snippet to `snippets/ruby-mode/definitions/mod`. This puts it under the `ruby-mode` directory.


AndreaCrotti:
Ah right thanks a lot I don't know how that got there in the first place :+1:

PR #114, opened 2016-01-16
Improvements to several elisp snippets

sshaw:
This makes them more flexible and puts them inline with their usage in the wild.

### DIFF
--- a/emacs-lisp-mode/add-hook
+++ b/emacs-lisp-mode/add-hook
@@ -1,6 +1,7 @@
+# -*- mode: snippet; -*-
 #contributor: Xah Lee (XahLee.org)
 #name: add-hook
 #key: add-hook
 #key: ah
 # --
-(add-hook ${1:hook} ${2:function})
+(add-hook '${1:name}-hook ${2:'${3:function}})$0
--- a/emacs-lisp-mode/lambda
+++ b/emacs-lisp-mode/lambda
@@ -1,6 +1,6 @@
+# -*- mode: snippet; -*-
 #contributor: Xah Lee (XahLee.org)
 #name: lambda
 #key: lam
 # --
-(lambda (${1:args})
-  $0)
+(lambda ($1) ${2:(interactive${3: "$4"}) }$0)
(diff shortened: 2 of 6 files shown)


emacs-lisp-mode/add-hook
```diff
@@ -1,6 +1,7 @@
-#contributor: Xah Lee (XahLee.org)
+# -*- mode: snippet; require-final-newline: nil -*-
+#contributor: Xah Lee (XahLee.org), Skye Shaw (https://github.com/sshaw)
```
AndreaCrotti: I'm not too sure about changing the contributor line..

I mean for me it's kind of pointless in general since Git keeps track of every possible change and we can always refer back to the contributors, but I don't think there is any snippet so far with multiple contributors declared in the snippet file itself..

sshaw: Modifications to the contributor line have been removed.

emacs-lisp-mode/add-hook
```diff
@@ -1,6 +1,7 @@
+# -*- mode: snippet; require-final-newline: nil -*-
```
AndreaCrotti: as we were discussing in https://github.com/AndreaCrotti/yasnippet-snippets/pull/117 this variable setting is not needed

AndreaCrotti:
And now as expected this conflicts with #117, could you please try to rebase again and see what changes should still be merged and were already done in the other branch?

Thanks and sorry for the complications

sshaw:
Hi, take a look now. 

My additions have better support for optional arguments and use more common defaults. Well, with the exception of `lambda` , maybe :smile:

AndreaCrotti:
Ok @sshaw I would merge it, the only thing I don't agree with is making all the names uppercased, because:
1. it's not really done in any other snippet
2. the actual name you need to write in is actually lowercase, so having it suggested uppercase might be confusing..

AndreaCrotti:
if you just lowercase them it's good to go

sshaw:
Okay, they're now lower case.

PR #116, opened 2016-01-20
Remove extraneous spaces at the end of common elisp snippets

thiderman:
I find myself using lots of these, but I always have to edit them to
remove the extra space they add. This patch removes all of those cases.


AndreaCrotti:
thanks not sure why there was the space in the first place..

PR #117, opened 2016-01-20
Major cleanup of emacs lisp snippets

thiderman:
The emacs lisp snippets in their current form are nigh unusable. This aims to fix that.

#### Changes
* Make the snippets actually use tab stops. Prior to these changes, more than half of the snippets had the form `(func ARGUMENT$0 &optional MORE-ARGUMENTS)`, presumably based on the function signatures found in the generated documentation. This adds more cruft than what makes it worth using the snippets, and the literal `&optional` is bordering on being a syntax error. The functions now have a form like `(func ${1:argument}${2: more-arguments})`, which is hopefully much more useful.
* As seen in the second example above, optional arguments have a preceding space. If the user don't want them, they can call `C-d` and everything including the leading space will be removed. If they want the argument, a press of space and then typing is sufficient.
* Snippets are updated to match the output as the modern indenting in `emacs-lisp-mode` produces.
* Remove duplicate snippets. `f` and `y-or-n-p` were duplicates of `format` and `yes-or-no-p` respectively. I updates their `#key:` values instead.
* Update lots of snippets to make them make more sense. Please refer to the individual `Update x` commits for more details.

---

These are a lot of changes, and some of them might be a bit opinionated. I'm super happy to change and update, so please give your feedback and I'll get right to it!

Relates to #116, since I started working on all these changes after I made that one. If that one feels superfluous and should have a home here, I can close that one and cherry pick that commit onto this branch instead.


AndreaCrotti:
Interestingly enough more or less at the same time there was another PR that touched the same files

https://github.com/AndreaCrotti/yasnippet-snippets/pull/114/files

They both look good, would be nice to get the best out of both maybe..
Could you please fix the merge conflicts first @thiderman and then I'll see how we can proceed with both PRs?
Thanks

thiderman:
Rebased!
Also I think the changes in #114 are all included inside the changes I made here. The only thing I didn't do was to add the `require-final-newline` flag. I also don't know what that does, haha.

Anything else I need to fix?

npostavs:
> The only thing I didn't do was to add the require-final-newline flag. I also don't know what that does

99% percent of the time you don't want a newline at the end of a snippet, so even if you have customized `require-final-newline` flag to non-nil, it should be nil for snippets. But having typed this, it occurs to me we can just set it to nil from the mode function, and there won't be any need for setting it in every snippet file.

npostavs:
> But having typed this, it occurs to me we can just set it to nil from the mode function, and there won't be any need for setting it in every snippet file.

Actually, it was already done years ago: https://github.com/capitaomorte/yasnippet/commit/5b48e2ec1517bf1216613956696c265c366af8e5#diff-62cbafd4b526f4c75b0362d03baae10bR446

AndreaCrotti:
Thanks a lot for this

thiderman:
Any time! It was fun! :dancer:

PR #118, opened 2016-01-22
go-mode: add a new snippet for parallel benchmark

kostya-sh:


AndreaCrotti:
Can't judge too much the implementation but looks good to me, thanks

PR #125, opened 2016-02-05
Update cout

simonpp:
match variable number of spaces after namespace

### DIFF
--- a/c++-mode/cout
+++ b/c++-mode/cout
@@ -4,5 +4,5 @@
 # key: cout
 # --
 `(progn (save-excursion) (goto-char (point-min)) (unless (re-search-forward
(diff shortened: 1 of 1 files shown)


AndreaCrotti:
mm is fine in general but @simonpp is it actually a normal thing to have more than one space there?
Why would you have more than one?

simonpp:
@AndreaCrotti  no it is not normal to have more than one space there, at least not for me. But it is legal syntax.

AndreaCrotti:
Ah I see, well to be honest if the only reason is that is valid syntax I prefer to keep it as it is..
If it's even a very unlikely but someone write code like that then I can merge it.

PR #130, opened 2016-03-22
use spaces instead of tabs in elixir mode

nimish-mehta:


AndreaCrotti:
Looks good to me, apparently 2 spaces is the suggested style anyway..

PR #135, opened 2016-04-13
Add two generic comment snippets for prog-mode

sh-ow:
I've added two helpful snippets for comments. These work in all programming modes, because they use comment-start and comment-end to determine the comment symbols.


npostavs:
Maybe you refactor code into functions that could go in `.yas-setup.el`? It's pretty much impossible to read as it is.

sh-ow:
Thanks for your feedback! I thought of these snippets as somehow 'write-only' code. But you are right: the readability could be much higher. I tried to improve this (see preceeding commit).

prog-mode/.yas-setup.el
```diff
@@ -1,2 +1,18 @@
 (defun yas-with-comment (str)
   (format "%s%s%s" comment-start str comment-end))
+
+(defun yas-get-comment-start ()
+  (substring comment-start 0
+			 (dotimes (i (length comment-start))
+			   (unless (or (eq (aref (reverse comment-start) i) nil)
```
npostavs: `reverse` doesn't work on strings in Emacs versions < 25.

prog-mode/.yas-setup.el
```diff
@@ -1,2 +1,18 @@
...
+						   (eq (aref (reverse comment-start) i) 32))
+				 (return (- (length comment-start) i))))))
```
npostavs: What does this function do? I think it's trying to trim whitespace from `comment-start`; starting from Emacs 24.4 you can use `string-trim`.

npostavs:
I'm still not really sure what these snippets do, but I would guess the existing `comment-box` and `comment-dwim` commands are better ways to do it.

prog-mode/.yas-setup.el
sh-ow: Thank you, I've overseen this. Should be fixed in 9b30fe4

prog-mode/.yas-setup.el
sh-ow: Yes, it is trimming whitespaces from `comment-start`. I didn't knew of `string-trim` but as it is only available since 24.4 I would like to leave the code as it is to support older versions.

sh-ow:
These snippets **do** basically the same like `comment-box` and `comment-dwim`. The difference is that these snippets fill the line until column 80 (and more if the comment is long enough) and therefore generate a more visible separator in code. I could have build some lisp around `comment-box` and `comment-dwin` to get this. But I do like the way with a snippet more than using these functions on a specific region after typing the comment. I think it does less interrupt the typing flow. This is why I thought these snippets could also be useful for others.

prog-mode/.yas-setup.el
npostavs: Okay, but at least name it something like `yas-trimmed-comment-start`, and put a docstring to explain what it's for.

npostavs: Oh, and your indentation is a bit wonky.

prog-mode/.yas-setup.el
```diff
@@ -1,2 +1,23 @@
...
+  (apply #'string
+	     (reverse
+	      (string-to-list str))))
+
+(defun yas-get-comment-start ()
+  (substring comment-start 0
+			 (dotimes (i (length comment-start))
+			   (unless (or (eq (aref (yas-string-reverse comment-start) i) nil)
+						   (eq (aref (yas-string-reverse comment-start) i) 32))
+				 (return (- (length comment-start) i))))))
```
npostavs: I think using `return` like that, only works if you have loaded `cl`. I think you would be better off counting down with `cl-loop` which would avoid the need for a string reverse. Or maybe do a regexp replace instead.

sh-ow: I've changed this to use the whitespace trim functions by Magnar Sveen (see e2409dd). These are quite small and by including them I make sure that this snippet also works on earlier versions of emacs.

prog-mode/.yas-setup.el
sh-ow: Renamed the functions (see e2409dd). Do you mean the indentation of .yas-setup.el? What is wonky about this?

prog-mode/.yas-setup.el
```diff
@@ -1,2 +1,37 @@
 (defun yas-with-comment (str)
   (format "%s%s%s" comment-start str comment-end))
+
+;; whitespace removing functions from Magnar Sveen ;;;;;;;;;;;;;;;;;;;;;;;;;;;;
```
npostavs: Is this an example of the `commentline` snippet output?

npostavs:
Okay, if you can just do <kbd>C-x h</kbd> <kbd>C-M-\\</kbd> in the `.yas-setup.el` buffer, it would look okay to me (though I personally wouldn't use these snippets), possibly the code remaining inline in the snippets themselves could still be simplified a bit, haven't really looked closely at it.

prog-mode/.yas-setup.el
sh-ow: Exactly. I've added a second semicolon on the beginning manually, to fullfill elisp commenting conventions.

prog-mode/.yas-setup.el
npostavs: The `(dotimes ...)` is indented too far to the right.

sh-ow:
Thanks again for your help! I've fixed indentation (4c0e664).

npostavs:
Hmm, it still looks wrong. Oh, you're indenting with tabs, and github uses 4 space tabs, whereas you have 8 space tabs. Lisp code is best indented with spaces. `(setq indent-tabs-mode nil)`

sh-ow:
Better with 00530f9?

npostavs:
Yup :)

AndreaCrotti:
I'll try it out, I'm just a bit concerned about the complexity of this solution, does it need to be so complicated?
Maybe some more functions could be extracted in the setup file btw, would like as little logic as possible in the actual snippets..

PR #136, opened 2016-04-21
Ruby: Replace all file-name underscores when creating module name.

patbl:
The previous implementation replaced only the first.


AndreaCrotti:
Thanks

PR #137, opened 2016-04-26
latex: added documentclass and moderncv snippets

toogley:
Snippet(s) for:

* **\documentclass** was missing, which is valuable because it is included in every LaTeX document.

* **moderncv commands** are also a valuable contributions IMHO, because e.g. `\cventry` has 5 Bracket pairs, which exact number i forget frequently.


AndreaCrotti:
Yes why not looks good

PR #138, opened 2016-04-26
Add snippet for C++11 classes

vedvyas:
Based on c++-mode/class. Allows for Rule of [0, All]. A choice between ";", " = default;", and " = delete;" is presented for each method. The methods and some of the optional keywords/specifiers are  exposed as fields that users can easily skip-and-clear.

References:
1. http://en.cppreference.com/w/cpp/language/rule_of_three#Rule_of_five
2. https://en.wikipedia.org/wiki/Rule_of_three_%28C%2B%2B_programming%29#Example_in_C.2B.2B
3. http://stackoverflow.com/a/4782927

### DIFF
--- a/c++-mode/class11
+++ b/c++-mode/class11
@@ -7,8 +7,7 @@
 # desc: Snippet for C++11 classes based on c++-mode/class. Allows for Rule of
 # [0, All]. A choice between ";", " = default;", and " = delete;" is presented
 # for each method. The methods and some of the optional keywords/specifiers are
-# exposed as fields that users can easily skip-and-clear. Note that the single
-# vs. differentiated copy/move assignment operators conflict.
+# exposed as fields that users can easily skip-and-clear.
 # Hackish query-replace-regexp to renumber non-mirror fields in the region
 # between public and protected (can use N as a field number in the snippet):
 # \${[0-9N]*:\([^\$]\) -> ${\,(+ 2 \#):\1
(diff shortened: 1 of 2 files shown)


c++-mode/class11
```diff
@@ -0,0 +1,44 @@
...
+# \${[0-9N]*:\([^\$]\) -> ${\,(+ 2 \#):\1
+# References:
+# 1. http://en.cppreference.com/w/cpp/language/rule_of_three#Rule_of_five
+# 2. https://en.wikipedia.org/wiki/Rule_of_three_%28C%2B%2B_programming%29#Example_in_C.2B.2B
+# 3. http://stackoverflow.com/a/4782927
+# --
+class ${1:Name}
+{
+public:
+${2:  //! Default constructor
```
AndreaCrotti: is it normal to put all these comments which I guess are more or less obvious?

Also if you are doing any more complex logic with Elisp could you please just put the code in some functions in the .yas-setup.el file in the directory of with the snippets?
Thanks a lot

vedvyas: These comment lines are brief descriptions for use with Doxygen. Here's an example of where they show up in Doxygen's HTML output:

doxygen_cls11_output

I agree that they are obvious but personally like them and sometimes there are more interesting comments that can be made about the methods :) Let me know if it is preferable to remove them or turn them into nested placeholders that users can edit or skip-and-clear.

I'll try to follow your suggestion about the logic as well. Thanks for the feedback.

AndreaCrotti:
Ok tried it out, there are a lot of placeholder but looks good to me, thanks!

vedvyas:
Thanks!

PR #140, opened 2016-04-29
Remove compiler warning

fice-t:
Makes sure yas-text is defined during byte compilation.


AndreaCrotti:
Yes I guess it doesn't harm :+1:

PR #142, opened 2016-05-06
complete template for article and moderncv class | safer bash settings

toogley:
Sorry for Commit, `ff5b2e1` I forgot to pull your changes in. Unfortunately, i currently don't know how to "reorder" that commit so that it doesn't appear in this PR, without rewriting all other Commits. Does this issue bother you?

### LaTeX

I don't know if it would be a valuable addition to provide full templates (specifically, I don't know whether Emacs has better options for automatically inserting templates when a new file is created). What do you think?

### Bash

source: http://redsymbol.net/articles/unofficial-bash-strict-mode/
Short Explanation:

    set -e               # fail if any command has non-zero exit status
    set -u               # fail if unbound variables are used
    set -o pipefail      # fail if command in a pipe isn't successful
    IFS=$'\n\t'          # wordsplitting will happen only on \n and \t, not on spaces
                         # => would otherwise break "My Song.mp3" or "Benjamin Franklin"

### DIFF
--- a/sh-mode/safe-bang
+++ b/sh-mode/safe-bang
@@ -1,0 +1,9 @@
+# -*- mode: snippet -*-
+# name: safer bash settings for scripts
+# key: s!
+# --
+#!/usr/bin/env bash
+set -euo pipefail
+IFS=$'\n\t'
+
+$0
--- a/latex-mode/article
+++ b/latex-mode/article
@@ -1,0 +1,11 @@
+# -*- mode: snippet -*-
+# name: full template of article class
+# key: article
+# --
+\documentclass[${1:options}]{article}
+
+\author{$3}
+
+\begin{document}
+$0
+\end{document}
--- a/latex-mode/moderncv
+++ b/latex-mode/moderncv
@@ -1,0 +1,15 @@
+# -*- mode: snippet -*-
+# name: full template of moderncv class
+# key: moderncv
+# --
+\documentclass[${1:options}]{moderncv}
+
+\firstname{$3}
+\familyname{$4}
+
+\moderncvstyle{${5:casual}}
+\moderncvcolor{${6:blue}}
+
+\begin{document}
+$0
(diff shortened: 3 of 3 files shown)


npostavs:
> Unfortunately, i currently don't know how to "reorder" that commit so that it doesn't appear in this PR, without rewriting all other Commits.

You could rebase, then `push -f` the result your branch.

toogley:
@npostavs  to cite from the git rebase manpage: 

> The todo list presented by --preserve-merges --interactive does not represent the topology of the revision graph. Editing commits and rewording their commit messages should work fine, but **attempts to reorder commits tend to produce counterintuitive results.**

npostavs:
In this case you don't want to preserve the merge anyway.

toogley:
Hm, that is at least a little improvement. Maybe i can improve that even further later.  @npostavs thanks for the tip.

npostavs:
You seems to have duplicated some other people's commits. Try an interactive rebase and mark those commits with `#` to get rid of them.

toogley:
Thanks, i didn't knew that feature.

sh-mode/safe-bang
```diff
@@ -0,0 +1,9 @@
+# -*- mode: snippet -*-
+# name: safer bash settings for scripts
+# key: s!
+# --
+#!/usr/bin/env bash
+set -euo pipefail
```
AndreaCrotti: this script is maybe not that generic but in any case it definitively doesn't harm and can be good to have..

AndreaCrotti:
Ok great thanks, in the future please do two different Pull requests (I mean splitting the bash changes from the latex changes)

sh-mode/safe-bang
toogley: Can you please explain what you meant by "generic"? I think every bash script does profit from such settings.

PR #144, opened 2016-05-24
add plantuml snippet

robbyoconnor:
Add a snippet for [plantuml](http://plantuml.com)

### DIFF
--- a/org-mode/uml
+++ b/org-mode/uml
@@ -1,0 +1,8 @@
+# -*- mode: snippet -*-
+#name : uml
+#key : uml
+#contributor : Robert O'Connor
+# --
+#+BEGIN_UML
+$1
+#+END_UML


robbyoconnor:
bump

AndreaCrotti:
That's cool but what I see from here
http://plantuml.com/emacs.html
that the actual syntax is not the one you used but something like

#+BEGIN_UML
  Alice -> Bob: Authentication Request
  Bob --> Alice: Authentication Response
#+END_UML

where is that syntax you used coming from?

robbyoconnor:
This isn't for org-mode -- but for actual plantuml files -- what you described is an org-mode codeblock

AndreaCrotti:
Well I said that because you placed the file in the org-mode directory, which means it's an org-mode snippet normally..

robbyoconnor:
OH doh -- i'll fix it

robbyoconnor:
Updated.

AndreaCrotti:
Mm I think it's still wrong @robbyoconnor 
it's in the UML directory which means it would be activated automatically on uml-mode, which does not exist as far as I can see (while plantuml-mode does).

On top of that if it's not org-mode specific what is the #+BEGIN #+END thing for?

I don't see anything like that here http://plantuml.com/sequence.html#Basic_examples

robbyoconnor:
Just fixed it

See http://plantuml.com/emacs.html

For org-babel   -- in short it's for org-mode UML blocks :)

AndreaCrotti:
Yes I know what org-babel is ;)
Ok thanks I can merge it now

PR #148, opened 2016-06-06
added python subclass snippet :snake:

marcanuy:
Added Python Inheritance snippet:

```
class DerivedClassName(BaseClassName):
    <statement-1>
    .
    .
    .
    <statement-N>
```

### DIFF
--- a/python-mode/scls
+++ b/python-mode/scls
@@ -1,0 +1,7 @@
+# -*- mode: snippet -*-
+# name: subclass
+# key: scls
+# group: object oriented
+# --
+class ${1:class}(${2:super-class}):
+    $0


AndreaCrotti:
I think we can actually merge this into the actual class maybe instead (making sure the superclass is optional).
Probably the subclasses defined in a project are normally way more than the classes without a super class anyway, what do you think?

marcanuy:
Agree, not sure how to make the second parameter optional though, I can think in two cases:

With the previous code and hitting `Ctrl+d` in the second parameter jumps to _$0_ leaving it empty:

```
class ${1:class}(${2:super-class}):
    $0
```
 
Or removing the parameter name from:

```
class ${1:class}(${2:}):
    $0
```

Is there any other method?

AndreaCrotti:
Well I think it's fine like this, in Python3 anyway we don't need to subclass object so the other case is quite common..

PR #151, opened 2016-06-24
Add org-mode code block snippet.

lzhoucs:
See [structure of code blocks](http://orgmode.org/manual/Structure-of-code-blocks.html#Structure-of-code-blocks) for details.

This is supposed to be one of the most commonly used `org-mode` snippets.

### DIFF
--- a/org-mode/code
+++ b/org-mode/code
@@ -1,0 +1,9 @@
+# -*- mode: snippet -*-
+# name: code
+# key: code_
+# contributor: Liang Zhou <lzhoucs@gmail.com>
+# --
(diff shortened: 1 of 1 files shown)


AndreaCrotti:
This snippet was actually there, but then I removed it because someone suggested 
<s TAB

which has the same behaviour with org-mode..

lzhoucs:
@AndreaCrotti , Cool! I didn't know that. It seems `org-mode` has [built in support for quite a few snippets](http://orgmode.org/manual/Easy-Templates.html). 

Thanks. Please close this one. :)

PR #152, opened 2016-07-03
Add include & Enumerable Snippet for Ruby

sshaw:


ruby-mode/Enum
```diff
@@ -0,0 +1,10 @@
+# -*- mode: snippet -*-
+# name: include Enumerable
+# key: Enum
+# group: collections
+# --
+include Enumerable
```
AndreaCrotti: I'm not a Ruby expert but is this a common scenario?

Is it normal to have "include Enumerable" and one "def each" just below on the same level, without classes or modules?

sshaw: Yes, it's common. Using `Enumerable` requires one define an `each` method. This is the case for many modules as is similar to [the snippet for `Comparable`](https://github.com/AndreaCrotti/yasnippet-snippets/blob/468635e2f52ac530d9a6bb29ee5ad2d7f651bd05/ruby-mode/Comp).

 Defining a class along with this snippet would make it less user friendly. 
There's already [a snippet for creating a class](https://github.com/AndreaCrotti/yasnippet-snippets/blob/468635e2f52ac530d9a6bb29ee5ad2d7f651bd05/ruby-mode/cls).

AndreaCrotti:
ok cool thanks

PR #155, opened 2016-08-01
Fix Emacs Lisp snippets being executable

strburst:
I was recently browsing the Emacs Lisp snippets when I was surprised to find most of them were executable (my ls colors show files that are executable in red). Since I don't think snippets count as exectuables, I've gone ahead and changed the permissions on those files.

With some `git bisect` magic, it seems this has been the case since the snippets were imported from the main yasnippet repo (5600362), with most subsequently added snippets having the correct permissions.

This is a fairly minor commit and doesn't have any changes to actual snippets.

You can verify that there are no more files that aren't supposed to be executable by inspecting the output of `find -executable -type f`.


AndreaCrotti:
Great thanks for finding out!

PR #156, opened 2016-08-03
Avoids warnings when compiling

npostavs:
I discovered in joaotavora/yasnippet#726 that `c++-mode/.yas-setup.el` is givng a warning when compiled.

```
* c++-mode/.yas-setup.el: Require yasnippet.
```

### DIFF
--- a/c++-mode/.yas-setup.el
+++ b/c++-mode/.yas-setup.el
@@ -1,3 +1,5 @@
+(require 'yasnippet)
+
 (defun yas-c++-class-name (str)
   "Search for a class name like `DerivedClass' in STR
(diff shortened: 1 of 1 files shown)


AndreaCrotti:
Ah good finding thanks @npostavs 
Any reason why it should not be added to all the .yas-setup.el that we have around?

npostavs:
They currently don't use yasnippet functions, so no warnings are triggered for them. It wouldn't hurt to add it anyway, though.

AndreaCrotti:
Yes ok I'll merge this one in the meanwhile anyway

PR #157, opened 2016-08-04
Added useful Elixir snippets

kwrooijen:
A few snippets that I use quite a lot during my Elixir development


AndreaCrotti:
I'm not an Elixir expert but looks pretty good to me, thanks @kwrooijen

PR #158, opened 2016-08-14
Add snippet for `function+docstring in Numpy convention` to python-mode

soupault:
see https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt


AndreaCrotti:
Nice one thanks

Issue #145, opened 2016-05-31
cc/c++-mode overloaded keys


ennokde:
Leaving this here as a proposition and a request for comments.

Currently when you hit tab on `inc` in cc-mode it offers you both includes, system and local and you have to pick one. I fixed this for myself by changing them to `incs` and `incl`.
Same thing for `f` in c++-mode which stands for a function declaration or a member function definition. I would suggest replacing the member function definition key with `mf`.

There is no shortage of (meaningful) abbreviations in c/c++-mode so I dont think there is a need for these collisions. If I am not alone with this thought I  get a patch ready of course.

nilsdeppe:
I agree with that there is no reason to have a collision in this case. I would love to see a patch so that by default `mf` is member function.

AndreaCrotti:
seems reasonable to me, sure a pull request is welcome, thanks

AndreaCrotti:
Closing this since it's already being taken care of in #159

PR #159, opened 2016-08-26
Fixing overloaded keys in cc/c++-mode

ennokde:
Finally fixing #145 
Sorry it took so long. Changed the keys as described in the issue.

### DIFF
--- a/c++-mode/function
+++ b/c++-mode/function
@@ -1,6 +1,6 @@
 # -*- mode: snippet -*-
 # name: function
-# key: f
+# key: mf
 # --
 ${1:type} ${2:Class}::${3:name}(${4:args})${5: const}
 {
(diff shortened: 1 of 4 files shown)


c++-mode/function
```diff
@@ -1,6 +1,6 @@
 # -*- mode: snippet -*-
 # name: function
-# key: f
+# key: mf
```
AndreaCrotti: what's the "m" for actually sorry?

ennokde: This snippet defines a Member Function for a class so mf is a better key for it to avoid the conflict with the fun_declaration snippet.
Speaking of it, the name should be changed to member_function aswell to avoid confusion with a free function. I added a commit to change the name too.

AndreaCrotti:
Ok good better now thanks

PR #163, opened 2016-10-11
Added algorithm-mnemonics for C++

ludwigpacifici:
Hello

Here are some snippets for standard C++ algorithm.  It is a set of abbreviations described in the repository [algorithm-mnemonics](https://github.com/tommybennett/algorithm-mnemonics). I did a port of it (in the repository [algorithm-mnemonics-emacs](https://github.com/ludwigpacifici/algorithm-mnemonics-emacs)) for Yasnippet.

I checked the names via `yas-describe-tables`. I saw no name collision for a buffer with `c++-mode`. The snippets are tested on my machine. Do you require more checks for this pull request?

Thanks

### DIFF
--- a/c++-mode/acl
+++ b/c++-mode/acl
@@ -1,7 +1,6 @@
 # -*- mode: snippet -*-
 # name: accumulate
 # key: acl
-# contributor: Tommy BENNETT and Ludwig PACIFICI <ludwig@lud.cc>
 # --
 auto sum = std::accumulate(std::begin(${1:container}), std::end($1), 0, [](int total, $2) {
   $3
--- a/c++-mode/acl
+++ b/c++-mode/acl
@@ -1,0 +1,9 @@
+# -*- mode: snippet -*-
+# name: accumulate
+# key: acl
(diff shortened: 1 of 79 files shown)


AndreaCrotti:
Hello @ludwigpacifici and thanks a lot.
I don't use C++ that much, are they all general and kind of common enough?

The contributor bit in each of the snippets might not be really useful since the author will still be preserved by git anyway..

ludwigpacifici:
I did a commit for the contributor part you mentioned.

These snippets are algorithms defined by the standard C++ library (some [examples](http://en.cppreference.com/w/cpp/algorithm)). They are "part of C++", meaning they are not snippets from a 3rd party library. For example the first snippet `acl` uses `std::accumulate` which defined [here](http://en.cppreference.com/w/cpp/algorithm/accumulate). Does it helps to reply to your question about generality of snippets?

Do you think it makes sense to put these snippets in a folder like `c++-algorithm`?

AndreaCrotti:
Yes ok thanks a lot, looks good to me

PR #167, opened 2016-11-10
Fix Elixir function placeholder text

lpil:
Elixir doesn't have methods, so I've updated this little snippet to be more accurate :)

### DIFF
--- a/elixir-mode/def
+++ b/elixir-mode/def
@@ -2,6 +2,6 @@
 # name: def
 # key: def
 # --
-def ${1:method}${2:(${3:args})} do
+def ${1:function}${2:(${3:args})} do
   $0
 end


elixir-mode/def
```diff
@@ -2,6 +2,6 @@
 # name: def
 # key: def
 # --
-def ${1:method}${2:(${3:args})} do
+def ${1:function}${2:(${3:args})} do
```
AndreaCrotti: if the "()" is always needed anyway I vote for @radik909 proposal.
If that's not the case leaving the option as it is now makes more sense imho..

lpil: The `()` is not always needed. Convention is to not put `()` if it is a zero arity function.

radik909: @lpil 
Got it!

AndreaCrotti:
All good thanks

lpil:
Thanks!

PR #168, opened 2016-11-17
Replace symlinks with .yas-parents file

npostavs:
```
When installing via the package.el, the symlinks in the tarball turn
into empty files.  To work around this, use .yas-parents files to
implement the linkage instead.
```

See joaotavora/yasnippet#748.


AndreaCrotti:
Ok sure thanks, I'll keep that in mind that using symlinks is not a good idea..

By the way I was also thinking to just publish this on Melpa as well.
If then yasnippet declares its dependency from it everything would be simpler right?

npostavs:
> If then yasnippet declares its dependency from it everything would be simpler right?

Maybe the dependency should go the other way around? It's possible to use yasnippet without using this particular set of snippets, after all.

PR #169, opened 2016-12-02
Add ipython embed snippet

halfcrazy:
Most time i use ipython embed to debug.

### DIFF
--- a/python-mode/embed
+++ b/python-mode/embed
@@ -1,6 +1,5 @@
 # -*- mode: snippet -*-
 # name: embed
 # key: embed
-# group: debug
 # --
 from IPython import embed; embed()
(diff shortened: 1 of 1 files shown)


AndreaCrotti:
thanks @halfcrazy 
I don't normally use this for debugging, but more ipdb or pdbpp.
I'm not sure that's what really ipython.embed is for exactly, so maybe I would just remove it from the "debug" group..

halfcrazy:
@AndreaCrotti It can open an interactive ipython shell when execute to that line. In the interactive shell, you have the current context and you can use ipython auto-complete feature to debug.

AndreaCrotti:
yes sure I know, I'm just saying that's not what it was for imho.
It's more to actual embed an interactive interpreter inside your application very easily, not really just for debugging.
But anyway I don't mind too much, but would be better to remove the "group: debug" line for me

halfcrazy:
@AndreaCrotti ping

AndreaCrotti:
Merged thanks

PR #170, opened 2016-12-02
Add a collection of lua snippets

halfcrazy:
Add some snippets for lua-mode, such as do, eif, eife, if, ife, iparis,
paris, repeat, while, require


AndreaCrotti:
Nice thanks!

PR #171, opened 2016-12-12
Python 'new' snippet uses dict as method argument

igroen:
The python 'new' snippet uses dict as method argument which is a python built-in type.
Using it as a method argument hides the built-in type 'dict' in the method body.


AndreaCrotti:
Nice catch thanks

PR #172, opened 2017-01-09
Updates to dix mode

unhammer:
I find these helpful at least.


AndreaCrotti:
I would not really know how to judge these but they look good to me.
Anyone else using dix-mode who can comment on it?

AndreaCrotti:
Ok merged thanks @unhammer

PR #173, opened 2017-01-29
Add some more regularly used snippets for C++ and C

108anup:
Added various snippets for vectors, long long int, some common headers, modified for loop in c-mode


c++-mode/F
```diff
@@ -0,0 +1,5 @@
+# -*- mode: snippet -*-
+# name: first
```
AndreaCrotti: not sure about first and second, I don't know how much they can be used but don't think it's worth to have a snippet for that, and definitively a single capital char is not really consistent with all the other snippets anyway..

c++-mode/pll
```diff
@@ -0,0 +1,5 @@
+# -*- mode: snippet -*-
+# name: pair<long long int,long long int>
+# key: pll
+# --
+pair<long long int,long long int> ${1:p};
```
AndreaCrotti: I don't think something like this is generic enough ..

AndreaCrotti:
I would get rid of first/second and make the vector/pair generic (unless for some reason they are always used with those types), but other than that it looks good.
Thanks

AndreaCrotti:
@108anup if you just resolve the couple of issues mentioned I'm happy to merge this

c++-mode/F
108anup: I added these for pair.first and pair.second, I will remove them if you wish.

c++-mode/pll
108anup: In competitive programming many people use pll for pair of long long with long long. That is why I thought this would be good.

108anup:
In many cases you almost always use vector with integers or long long, but for a generic version we could keep vec [Tab] which gives option for type as well as name, keeping the special case vi and vll

c++-mode/F
AndreaCrotti: Yes well also because (I don't know why C++) but with company-mode and most languages you will get autocompletion anyway right?

I would get rid of this if possible..

c++-mode/pll
AndreaCrotti: Ok I see, well it still sounds quite like a corner case, I'm sure competitive programmers have their own crazy snippets anyway to go as fast as possible :)

AndreaCrotti:
@108anup if we could remove at least the first (as mentioned above) I would be fine in merging this.
Thanks!

beta1440:
Configuring code completion like `company-clang` or `ac-clang` will provide additional functionality. @108anup you might be better off creating a repository just for competitive programming snippets. That way, you can add as many snippets for competitive code as you want.

AndreaCrotti:
Yes good point @beta1440 
I'm happy to merge the more generic snippets though if you make another PR @108anup 
I'll close this PR in the meanwhile since it conflicts anyway, thanks

PR #175, opened 2017-02-17
Use minimum indentation for python docstring

npostavs:
Fixes joaotavora/yasnippet#750.

```
* python-mode/function_docstring: Reduce indentation to 1, because
Emacs 25+ doesn't dedent docstrings.
```


AndreaCrotti:
Cool thanks

PR #176, opened 2017-02-26
Remove side-effecting expression inside backquotes

npostavs:
```
* c++-mode/cout: Remove unneeded `save-excursion'.  Yasnippet already
uses `save-excursion' and the call here had no arguments which made it
a nop.
* cc-mode/case: Indent the first line by binding
`yas-also-auto-indent-first-line' instead of calling `indent-region'.
```

See also http://emacs.stackexchange.com/questions/30977/why-does-indent-region-insert-a-t-into-my-buffer

Actually, the `cout` isn't really a problem, I just happened to see it while searching for other problematic snippets.

### DIFF
--- a/c++-mode/cout
+++ b/c++-mode/cout
@@ -3,6 +3,6 @@
 # name: cout
 # key: cout
 # --
-`(progn (save-excursion) (goto-char (point-min)) (unless (re-search-forward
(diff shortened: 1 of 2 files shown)


AndreaCrotti:
Thanks @npostavs

PR #179, opened 2017-03-29
Add applescript snippets from apples-mode

suzume:
apples-mode had snippets that were not in this repository:
https://github.com/tequilasunset/apples-mode

### DIFF
--- a/apples-mode/.yas-parents
+++ b/apples-mode/.yas-parents
@@ -1,0 +1,1 @@
+prog-mode
--- a/apples-mode/comment
+++ b/apples-mode/comment
@@ -1,0 +1,6 @@
+# -*- mode: snippet -*-
+# name: comment
+# --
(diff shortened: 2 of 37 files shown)


AndreaCrotti:
Should generally ask permission to the author before taking snippets from other repositories @suzume 
Did you ask already maybe?

suzume:
The original repository has been inactive for 6 years, right after the original code has been committed. I've sent a pull request to make the code compatible with recent versions of OSX but have not yet received any answer. I've also sent a mail to the address documented in the source, no reaction too.

suzume:
@AndreaCrotti is there anything else I need to do to have that request considered ?

AndreaCrotti:
Here I am sorry, yes it's fine I think we it's fine to copy them over in this particular case.
Thanks!

PR #182, opened 2017-04-04
* cc-mode/once: Don't assume buffer is visiting a file.

### LINKED ISSUE #181: cc-mode once does not expand correctly
Hello. Thank you for the collection of snippets.
I am having a bug with cc-mode once: when I try to complete the snippet with C-S-i (see configuration below), I get the following:
```
#ifndefY_H}
#define 



#endif /*  */
```
And I see the following line in the *Messages* buffer
`yas--snippet-parse-create: Wrong type argument: char-or-string-p, (stringp nil)`

I looked into the snippet, and tried to debug. So, I went to *scratch*, and I get an error with this:
`(file-name-sans-extension (buffer-file-name))`
```
Debugger entered--Lisp error: (wrong-type-argument stringp nil)
  file-name-nondirectory(nil)
  file-name-sans-extension(nil)
  eval((file-name-sans-extension (buffer-file-name)) nil)
  eval-last-sexp-1(nil)
  eval-last-sexp(nil)
  call-interactively(eval-last-sexp nil nil)
  command-execute(eval-last-sexp)
```
Which seems to be relevant only when the buffer is not saved. When the file is saved, the name does appear next to `ifndef` (without space in between), but it is not expanded in the rest of the snippet, and hitting tab does not move the mark to the next tag.

I also have this in *Warnings*
```
Warning (yasnippet): Couldn't find: "[[:space:]
]*#ifndef[[:space:]
]*TEST_H}[[:space:]
]*#define[[:space:]
]*#endif[[:space:]
]*/\\*[[:space:]
]*\\*/[[:space:]
]*\\(\\)[[:space:]
]*#endif[[:space:]
]*/\\*[[:space:]
]*\\*/"
```

If I knew how to fix it, I would propose a solution, but I am really not good with lisp. I tried with an actual file and a new c-mode buffer. Also, I don't know if this has to do with the snippet alone or yasnippet as a whole (may be related to https://github.com/joaotavora/yasnippet/issues/784).

My configuration (running emacs -Q):
emacs GNU Emacs 24.4.1 (x86_64-pc-linux-gnu, GTK+ Version 3.4.2)
yasnippet-20170326.1030 from melpa

.emacs (init file)
```
(require 'package)
(package-initialize)
(require 'yasnippet)
(define-key yas-minor-mode-map (kbd "<tab>") nil)
(define-key yas-minor-mode-map (kbd "TAB") nil)
(define-key yas-minor-mode-map (kbd "C-i") nil)
(define-key yas-minor-mode-map (kbd "C-S-i") 'yas-expand)

(add-hook 'prog-mode-hook 'yas-minor-mode)

(require 'function-args)
(add-hook 'c-mode-common-hook (lambda ()
                                (function-args-mode) ; Find definitions in current buffer
                                (setq tab-width 4) ; set tab width to 4
                                (auto-fill-mode) ; break long lines
                                (yas-minor-mode) ; activate yasnippet
                                ))
```

npostavs:
Fixes #181 (part 1 that is, part 2 is fixed in joaotavora/yasnippet#798).

### DIFF
--- a/cc-mode/once
+++ b/cc-mode/once
@@ -1,7 +1,7 @@
 #name : #ifndef XXX; #define XXX; #endif
 # key: once
 # --
-#ifndef ${1:`(upcase (file-name-nondirectory (file-name-sans-extension (buffer-file-name))))`_H}
+#ifndef ${1:`(upcase (file-name-nondirectory (file-name-sans-extension (or (buffer-file-name) ""))))`_H}
 #define $1
 
 $0


AndreaCrotti:
Great thanks @npostavs

PR #183, opened 2017-04-11
lua mode "paris" -> "pairs" typos 

jameslintaylor:
I can only assume that the naming of (i)pairs snippets as (i)pa*ri*s were simple typos. This amends them 😄


AndreaCrotti:
Yes ok thanks

PR #184, opened 2017-04-13
add begin align snippet

arthurcgusmao:
Added new \begin{align} snippet for org-mode with key "begal_".


AndreaCrotti:
Not too sure about this one, it's probably not as commonly used and the key is quite hard to remember..

I suggest to keep it in your personal snippets, thanks anyway!

PR #185, opened 2017-04-16
Modifying Python setUp and tearDown methods unittests

k-schmidt:
This merge serves to provide a better template for Python unittest classes. A user incrementally chooses whether to modify the setup and teardown methods themselves or leave as 'pass'.

I find in code that even if the testing suite does not utilize setup and teardown methods it is still helpful to leave them there as 'pass' so to tell any reader that this testing suite does not use or require those methods.

It will also perhaps enforce better testing structure down the line for those who might not know about these features of the library.

### DIFF
--- a/python-mode/test_class
+++ b/python-mode/test_class
@@ -7,16 +7,16 @@ class Test${1:toTest}(${2:unittest.TestCase}):
 
     @classmethod
     def setUpClass(cls):
-        pass
+        ${3:pass}
 
     @classmethod
     def tearDownClass(cls):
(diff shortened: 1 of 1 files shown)


AndreaCrotti:
Uhm I'm not too sure about this one..
It does force everyone to delete manually all these methods, even if you don't actually need them.
I personally don't use setup/teardowns very often, and I think it never happened I had to implement all four of them..
It's nice in theory to help people learning about new things, but it's not rally the purpose of snippets imho..

PR #187, opened 2017-04-28
make module name in testmod editable field

tomcraven:


AndreaCrotti:
Looks reasonable, thanks

PR #188, opened 2017-04-30
Update function template to prevent additional space insertion.

varunpatro:


js-mode/function
```diff
@@ -2,6 +2,6 @@
 # name: function
 # key: f
 # --
-function${1: ${2:name}}(${3:arg}) {
+function ${1:${2:name}}(${3:arg}) {
```
npostavs: You should get rid of the doubled field as well

AndreaCrotti:
if you address what @npostavs mentioned @varunpatro I'll merge this, thanks!

varunpatro:
Had to create another PR #189. I'll close this one :)

PR #190, opened 2017-05-11
Make c++-mode access specifier snippets indent the first line

npostavs:
```
The first line of the expansion usually needs to be unindented
relative to the current offset.
* c++-mode/private:
* c++-mode/protected:
* c++-mode/public: Bind yas-also-auto-indent-first-line to t.
```


AndreaCrotti:
Thanks @npostavs !

PR #191, opened 2017-06-16
Disable recursive snippets in Clojure namespace snippets

jgertm:
Previous to this, typing `clojure.test` and then tabbing to go to the next field
would result in an expansion of test.


AndreaCrotti:
Great thanks a lot!

PR #193, opened 2017-06-29
Add python/ipython snippets for org-mode

daviderestivo:
Ciao Andrea,

I've created a couple of snippets for python and ipython (org-mode). Let me know if you think it might be a good idea to merge it to your repo.

Have a nice day and regards,
Davide

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -1,4 +1,5 @@
 Please add here your name in alphabetical order (first name) if you contributed with
 
 - Andrea Crotti
+- Davide Restivo
 - James Ferguson
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -1,5 +1,4 @@
 Please add here your name in alphabetical order (first name) if you contributed with
 
 - Andrea Crotti
-- Davide Restivo
 - James Ferguson
--- a/org-mode/ipython
+++ b/org-mode/ipython
@@ -2,6 +2,6 @@
 # name: ipython
 # key: ipy_
 # --
-#+BEGIN_SRC ipython :session ${1:session01} :file ${2:$$(progn (setq temporary-file-directory "./") (concat temporary-file-directory "ipython-" (make-temp-name "") ".png"))} :exports ${3:both}
+#+BEGIN_SRC ipython :session ${1:session01} :file ${2:$$(concat (make-temp-name "./ipython-") ".png")} :exports ${3:both}
 $0
 #+END_SRC
--- a/org-mode/ipython
+++ b/org-mode/ipython
@@ -1,0 +1,7 @@
+# -*- mode: snippet -*-
+# name: ipython
+# key: ipy_
+# --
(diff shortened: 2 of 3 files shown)


AndreaCrotti:
It's a nice idea, the only thing is that I'm not sure if yasnippet plays very well with org-mode, since it re-binds so many things.
I actually use `<s TAB` there to create src snippets, and with my config yasnippet doesn't work at all.

Having said that maybe there is still value in having some more complicated snippets, but about ipython in particular what do you use it with org-mode for?

Thanks for your contribution!

org-mode/ipython
```diff
@@ -0,0 +1,7 @@
+# -*- mode: snippet -*-
+# name: ipython
+# key: ipy_
+# --
+#+BEGIN_SRC ipython :session ${1:session01} :file ${2:$$(progn (setq temporary-file-directory "./") (concat temporary-file-directory "ipython-" (make-temp-name "") ".png"))} :exports ${3:both}
```
npostavs: I think the code in `${2:...}` would be equivalent to just `(concat (make-temp-name "./ipython-") ".png")`

daviderestivo: Sure, they are equivalent but I found the explicit reference to the temporary directory more clear. I have no issue if you wanna change it.

Thanks in advance and regards,
Davide

daviderestivo:
Ciao Andrea,

I'm using yasnippet in org mode and to be honest i'm very happy with it. Currently I use ob-ipython (ORG Babel module) in order to use a buffer as a jupiter notebook. To be honest the fact that you can manage the code, comment and task list has a big advantage.

This is a screenshot from the ob-ipython page: https://github.com/gregsexton/ob-ipython/blob/master/screenshot.jpg)

Have a nice day and regards,
Davide

org-mode/ipython
npostavs: Well, I would definitely say to at least not permanently modify `temporary-file-directory` as that could break other Emacs code (i.e., you should use `let` instead of `setq`).

My personal opinion is that the shorter version is clearer, as you are not really making use of `temporary-file-directory` anyway so it ends up just being a distraction.

daviderestivo: Fine with me ... go for the short version.

AndreaCrotti: I think I agree with @npostavs and just to clarify @daviderestivo normally it's the person opening the PR that should make the change, not the reviewer.

If you change that I'm happy to merge though, thanks

CONTRIBUTORS.txt
```diff
@@ -1,4 +1,5 @@
 Please add here your name in alphabetical order (first name) if you contributed with
 
 - Andrea Crotti
+- Davide Restivo
```
AndreaCrotti: This file was actually deleted entirely since it's kind of pointless, only a couple of people were there and there are way way more contributors all already tracked by git/github anyway.
So please when you resolve the conflict just ignore this change

daviderestivo:
Hi Andrea,

I've done the changes as discussed. Please let me know if it's ok now to merge.

Have a nice weekend and regards,
Davide

org-mode/ipython
daviderestivo: @AndreaCrotti: I did the changes as requested ... is there anything missing to merge?

Thanks,
Davide

AndreaCrotti:
Thanks @daviderestivo

daviderestivo:
Thanks to you @AndreaCrotti

PR #194, opened 2017-07-10
Update python-mode snippet

zhenzong:


python-mode/str
```diff
@@ -4,4 +4,6 @@
 # group: dunder methods
 # --
 def __str__(self):
-    $0
\ No newline at end of file
+    $0
+
+__repr__ = __str__
```
AndreaCrotti: mm not sure I agree with this one, repr is often not the same as str..

python-mode/doc
```diff
@@ -2,5 +2,6 @@
 # name: doc
 # key: d
 # --
-"""$0
+"""
+$0
```
AndreaCrotti: This also is not necessarily always the right thing to do I think..

python-mode/str
zhenzong: Ok, I will discard this modify. Thank for reply.

python-mode/doc
zhenzong: Your are right. I check [pep](https://www.python.org/dev/peps/pep-0257/) again, my fault.

AndreaCrotti:
Ok great thanks

Issue #195, opened 2017-07-14
cc-mode for: i++ -> ++i


nico202:
https://stackoverflow.com/questions/4261708/i-or-i-in-for-loops#4261743

It should be better to use ++i

PR #197, opened 2017-07-14
cc-mode: i++ -> ++i

nico202:
#195


AndreaCrotti:
Yes thanks I vaguely remember something about that too, makes sense

PR #199, opened 2017-07-21
cc-mode: Improve readability of main

beta1440:
Each array in argv always contains argc amount of strings. Specify it explicitly
for better readability.

### DIFF
--- a/cc-mode/main
+++ b/cc-mode/main
@@ -2,7 +2,7 @@
 # name: main
 # key: main
 # --
-int main(${1:int argc, char *argv[]})
+int main(${1:int argc, char *argv[argc]})
 {
     $0
     return 0;


AndreaCrotti:
Yes ok why not, thanks!

cc-mode/main
```diff
@@ -2,7 +2,7 @@
 # name: main
 # key: main
 # --
-int main(${1:int argc, char *argv[]})
+int main(${1:int argc, char *argv[argc]})
```
npostavs: I think this syntax is not legal for all versions of C, perhaps only since C99? (which could be okay, but it's something to consider at least).

AndreaCrotti: ah ok I see, I tried it out quickly on mine and didn't give any warnings so I assumed it was ok.
I think if it's actually C99 we are fine (that was 18 years ago), but if anyone that knows more C than me can confirm it would be great.
Thanks @npostavs

npostavs: - Emacs requires C99 starting from version 25.1. 
- gcc gives a warning with `-std=c89 -pedantic`, but not with `-std=c89`. I think that means this feature was available long before C99, so probably it's fine.
```
$ make -B foo.o CFLAGS='-std=c89 -pedantic'
cc -std=c89 -pedantic   -c -o foo.o foo.c
foo.c:1:1: warning: ISO C90 forbids variable length array 'argv' [-Wvla]
 int main(int argc, const char * argv[argc]) {
 ^~~
```

AndreaCrotti: Ok thanks I think it's fine then

PR #200, opened 2017-07-22
Add remote_pry snippets for ruby-mode

mikhail-krainik:
Hey folks. I'd like to add a new snippet in ruby-mode. 
It's provide remotely debug by pry (using DRb)
I hope it will be usable.
```
require 'pry-remote'; binding.remote_pry
```

### DIFF
--- a/ruby-mode/rpry
+++ b/ruby-mode/rpry
@@ -1,0 +1,5 @@
+# -*- mode: snippet -*-
+# name: binding.pry_remote
+# key: rpry
+# --
+require 'pry-remote'; binding.remote_pry


mikhail-krainik:
Hey folks. I'd like to add a new snippet in ruby-mode. 
It's provide remotely debug by pry (using DRb)
I hope it will be usable.

AndreaCrotti:
Don't use ruby much, is that something commonly used enough to be added as snippet you think @mikhail-krainik ?

mikhail-krainik:
I think remote_pry is very usable at the level other snippets in ruby-mode because debugging is boring things and pry or remote_pry were typing high frequency than some less usable snippets. And a same for testing. I think my snippet will be usable. I use that and it was not enough for me. Actually I think about provide snippets for rspec-mode for fast typing next words "is_experted", "let(:$(var)) { $(0) }" etc.. if my enthusiasm will be received positively.

AndreaCrotti:
ok great thanks @mikhail-krainik

PR #202, opened 2017-07-28
Add PHP basic syntax (statement) for php-mode

zonuexe:
These are snippets of PHP's basic syntax.

### DIFF
--- a/snippets/php-mode/.yas-setup.el
+++ b/snippets/php-mode/.yas-setup.el
@@ -1,0 +1,16 @@
+;;; .yas-setup.el --- Yasnippet helper functions for PHP snippets
+
+;;; Commentary:
+
+;;; Code:
+(require 'yasnippet)
+
+(defun yas-php-get-class-name-by-file-name ()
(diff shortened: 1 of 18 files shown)


zonuexe:
@AndreaCrotti Could you review this PR?

AndreaCrotti:
Here I am @zonuexe 
The snippets look really good, the only thing I would say is that you don't need to repeat the same code all over the place
(like `(file-name-nondirectory (file-name-sans-extension (or (buffer-file-name) (buffer-name (current-buffer)))))` )

And you can put the common stuff in a `.yas-setup.el` file in the php-mode directory.
See an example of how it's done for Python here.

 https://github.com/AndreaCrotti/yasnippet-snippets/blob/master/snippets/python-mode/.yas-setup.el

Other than that happy to merge it, thanks a lot!

zonuexe:
@AndreaCrotti Thanks your review. I've been fixed it.

AndreaCrotti:
Ok great thanks!

PR #205, opened 2017-08-21
Add snippets for php-mode

zonuexe:
Add magic methods, popular [PHPDoc tags](https://phpdoc.org/docs/latest/internals/tags.html) and convenience idioms for debugging.

### DIFF
--- a/snippets/php-mode/license-doc
+++ b/snippets/php-mode/license-doc
@@ -1,0 +1,8 @@
+# -*- mode: snippet -*-
+# contributor: USAMI Kenta <tadsan@zonu.me>
+# name: PHPDoc @license
+# key: *license
+# group: annotation
+# --
+* @license ${1:https://www.gnu.org/licenses/gpl-3.0.html} ${2:GPL-3.0}${0:
+*}
--- a/snippets/php-mode/__call
+++ b/snippets/php-mode/__call
@@ -1,0 +1,10 @@
+# -*- mode: snippet -*-
+# contributor: USAMI Kenta <tadsan@zonu.me>
+# name: __call()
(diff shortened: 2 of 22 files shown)


AndreaCrotti:
Thanks a lot @zonuexe I'll try to have a deeper look next week!

snippets/php-mode/license-doc
```diff
@@ -0,0 +1,8 @@
+# -*- mode: snippet -*-
+# contributor: USAMI Kenta <tadsan@zonu.me>
+# name: PHPDoc @license
+# key: *license
```
AndreaCrotti: Any reason to prepend all license snippets with *?
It doesn't make it easy to remember the snippet imho..

AndreaCrotti:
I'm not sure about all the keys prefixed by "*", also because I don't see a pattern for which ones should be and which ones are not prefixed.

If we can clarify/fix that @zonuexe then I'm happy to merge.

zonuexe:
@AndreaCrotti 
It can be explained by the place where @AndreaCrotti license is written.
Snippets with "Doc" in their name are carefully designed based on the actual input process.

## Manual typing

yasnippet-php-freehand

## Use yasnippet

yasnippet-php-yasnippet

------

`@license` has to be written in DocComment, so it is always entered with `* @`.
I believe that `*` is reasonable and best because of the connectivity of the input with other tags for the PHPDoc tag prefix.

zonuexe:
I found my typo with this video ;-)

AndreaCrotti:
Yes ok thanks @zonuexe makes sense

zonuexe:
Thanks!

PR #207, opened 2017-08-22
Correct terraform 'resource' snippet

timanin:
Type definition was missing. As the official Terraform documentation states:

> The resource block defines a resource that exists within the infrastructure. A resource might be a physical component such as an EC2 instance, or it can be a logical resource such as a Heroku application.

From: https://www.terraform.io/intro/getting-started/build.html#configuration


AndreaCrotti:
Ah great thanks nicely spotted

PR #209, opened 2017-08-28
Fix detection of current file name

### LINKED ISSUE #208: Improve detection of current file in yasnippet-snippets.el
yasnippet-snippets.el uses `(or (buffer-file-name) load-file-name)` to find its own filename, cf. https://github.com/AndreaCrotti/yasnippet-snippets/blob/master/yasnippet-snippets.el#L27. However, that's incorrect if e.g. the user does `M-x load-library yasnippet-snippets` from a file-visiting buffer. I'd suggest to either call `f-this-file` from f.el or reproduce its logic.

phst:
Fixes #208

### DIFF
--- a/yasnippet-snippets.el
+++ b/yasnippet-snippets.el
@@ -24,8 +24,14 @@
 
 (require 'yasnippet)
 
-(setq yasnippet-snippets-dir (file-name-directory (or (buffer-file-name)
-						      load-file-name)))
+(setq yasnippet-snippets-dir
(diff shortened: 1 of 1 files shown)


AndreaCrotti:
Ah yes you're right @phst.
I took it from other snippets and didn't notice since it just auto loads with `yasnippet` with this configuration, but it's true it fails in the scenario you mentioned.
Thanks a lot

PR #210, opened 2017-09-01
Update python mode metaclass snippet

zhenzong:


AndreaCrotti:
Nice thanks @zhenzong

PR #212, opened 2017-09-07
Fix path for csharp-mode/main

thblt:
See commit.  csharp-mode/main was outside `/snippets`


AndreaCrotti:
Ah I see it was probably merged while I was doing the switch and didn't get moved over.
Thanks for noticing @thblt

PR #213, opened 2017-09-07
create python snippet bang #!

avelino:


AndreaCrotti:
Sure why not, we have one already but only for bash

PR #216, opened 2017-09-13
use "forEach" instead of "each"

ybiquitous:
"forEach" is standard API, but "each" is non-standard.

See below:

- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach
- https://developer.mozilla.org/en-US/docs/Web/API/NodeList/forEach


AndreaCrotti:
Ah cool thanks @ybiquitous

Issue #134, opened 2016-04-12
add es6 snippets


transtone:
please add js's es6 snippets like vsc:
https://marketplace.visualstudio.com/items?itemName=xabikos.JavaScriptSnippets

AndreaCrotti:
I would probably not just copy things from another place since they have to work well with the existing snippets..
And I'm not really a JS developer so would not do it myself since I can't judge.
PR welcome of course ;)

AndreaCrotti:
Quite a few things where merged already now related to this issue so I think it's fair to assume it can be closed now

PR #218, opened 2017-09-19
rewrite 'class' and 'constructor' for ES2015 syntax in js-mode

ybiquitous:
partial fix for #134

For details, see [Classes - JavaScript | MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes).

### DIFF
--- a/snippets/js-mode/class
+++ b/snippets/js-mode/class
@@ -1,8 +1,12 @@
-# -*- mode: snippet -*-
-#name : Class
+# -*- mode: snippet; require-final-newline: nil -*-
+# name: class
+# key: cls
 # --
-var ${1:name} = new Class({
-  initialize: function($2) {
-    $0
-  }
-});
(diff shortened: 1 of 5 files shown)


ybiquitous:
This PR breaks compatibility, but ES2015 (ES6) becomes very familiar recently, so I think that it will not cause big problems.

What do you think about this? Please I hope any feedbacks!

AndreaCrotti:
Well I'm not sure it's worth to break backward compatibility @ybiquitous 
I'm not an expert but lots of people are still not writing ES6 all the time right?
Could we not simply add another snippet and leave the original one as it is?

Another option is to have both snippets there with the same key, in that case yasnippet simply provides you with the option to choose which one you need.

ybiquitous:
@AndreaCrotti 
I pushed 5c214f9 to keep compatibility. Please review again 🙏

AndreaCrotti:
Cool thanks @ybiquitous

PR #219, opened 2017-09-19
add snippets about ES2015 module syntax (import, export)

ybiquitous:
For details, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import

partial fix for #134

### DIFF
--- a/snippets/js-mode/import
+++ b/snippets/js-mode/import
@@ -1,0 +1,5 @@
+# -*- mode: snippet; require-final-newline: nil -*-
+# name: import
+# key: imp
+# --
+import ${1:module}${2:, { ${3:member} }} from '$1';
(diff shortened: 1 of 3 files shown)


snippets/js-mode/import
```diff
@@ -0,0 +1,5 @@
+# -*- mode: snippet; require-final-newline: nil -*-
+# name: import
+# key: imp
+# --
+import ${1:module}${2:, { ${3:member} }} from '$1';
```
AndreaCrotti: I'm not an expert but the `from $1` seems wrong to me, because from
https://developer.mozilla.org/es/docs/Web/JavaScript/Referencia/Sentencias/import
the $module and the actual name of the file would never be exactly the same (at least the .js at the end).

ybiquitous: @AndreaCrotti 
Surely, you're right! I pushed c53d634, so please review again 🙏

ybiquitous:
@AndreaCrotti 

> Could you also add a "group: es6" in all the snippets that add ES6 syntax?

Good idea! I pushed 8e6ffbf 👍

PR #220, opened 2017-09-21
Add powershell-mode with basic function snippets

mehak:
Hi Andrea,

I added a couple snippets that make writting PowerShell functions easier.  These are the first snippets I have written and I am not sure if I should be doing things a little different (like the key on comment-based-help being cbh).  If you have any suggestions or quality control issues, please let me know.

I just found out about Yasnippet and it has been great, thank you!

Regards,
Nate


AndreaCrotti:
I really know nothing about powershell and don't have a way to try them out, but the snippets themselves look good, so it's fine for me.

PR #221, opened 2017-09-24
Put $0 before comment-end

npostavs:
Fixes joaotavora/yasnippet#854.
```
* snippets/prog-mode/fixme:
* snippets/prog-mode/todo:
* snippets/prog-mode/xxx: Use comment-start and comment-end directly
instead of `yas-with-comment` so that $0 can be inserted before comment-end.
```

### DIFF
--- a/snippets/prog-mode/.yas-setup.el
+++ b/snippets/prog-mode/.yas-setup.el
@@ -1,8 +1,5 @@
 (require 'yasnippet)
 
-(defun yas-with-comment (str)
-  (format "%s%s%s" comment-start str comment-end))
-
 ;; whitespace removing functions from Magnar Sveen ;;;;;;;;;;;;;;;;;;;;;;;;;;;;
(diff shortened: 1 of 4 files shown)


AndreaCrotti:
Thanks @npostavs 
Would be actually nice to keep using this function instead though

https://github.com/AndreaCrotti/yasnippet-snippets/blob/master/snippets/prog-mode/.yas-setup.el#L3

It's actually only used in these three instances so if we don't use it anymore we can just delete it instead.
Thanks!

npostavs:
> Would be actually nice to keep using this function though, is it not easier to just modify that?

Unfortunately, there's no way for a function to return a field like `$0`.

> It's actually only used in these three instances so if we don't use it anymore we can just delete it instead.

Okay, done.

AndreaCrotti:
Ok cool thanks!

PR #222, opened 2017-09-26
Update cc-mode/main

Davidbrcz:
The current cc-mode/main  snippet does not compile out of the box with gcc or g++.
I reverted it to a more standard code which does not use variable length array with its arguments

### DIFF
--- a/snippets/cc-mode/main
+++ b/snippets/cc-mode/main
@@ -2,7 +2,7 @@
 # name: main
 # key: main
 # --
-int main(${1:int argc, char *argv[argc]})
+int main(${1:int argc, char *argv[]})
 {
     $0
     return 0;


AndreaCrotti:
This was actually done through:
https://github.com/AndreaCrotti/yasnippet-snippets/pull/199
I'm surprised it doesn't compile, but are you using something older than c89?
I'll have a try myself

Davidbrcz:
I'm using only C++, it works when compiling C code with clang or gcc.
It fails with g++ but works with clang++
```
david:~/code $ cat main.cpp                                                                                                                                                        
int main(int argc, char *argv[argc])
{
 
  return 0;
}
david:~/code $ g++ main.cpp -o main                                                                                                                                                
main.cpp:1:26: error: expected ‘,’ or ‘...’ before ‘argv’
 int main(int argc, char *argv[argc])
                          ^
main.cpp:1:5: warning: second argument of ‘int main(int, char*)’ should be ‘char **’ [-Wmain]
 int main(int argc, char *argv[argc])
     ^
david:~/code $ g++-7 -std=c++14 main.cpp -o main                                                                                                                                   
main.cpp:1:26: error: expected ‘,’ or ‘...’ before ‘argv’
 int main(int argc, char *argv[argc])
                          ^~~~
main.cpp:1:5: warning: second argument of ‘int main(int, char*)’ should be ‘char **’ [-Wmain]
 int main(int argc, char *argv[argc])
     ^~~~
david:~/code $ clang++ main.cpp -o main                                                                                                                                            
david:~/code $ clang++ -std=c++14 main.cpp -o main       
david:~/code $
```

npostavs:
Perhaps we should move this one to `c-mode`, and give the non-vla version to `c++-mode`. Do C# and nesC (the other children of cc-mode) even use `main`?

Davidbrcz:
After a quick tour on google, C# has a main but not like the one in the snippet. I could not find something conclusive for nesC (It seems to be a decade old project not really active). 

All I can be sure is that the updated snippet works with both C and C++.

AndreaCrotti:
Mm pff I think I should simply not have merged the other change that added `argc` in the first place, the benefit was probably negligible anyway.
I'll merge this revert for now, and maybe we can do something a bit smarter in the future.

PR #225, opened 2017-09-27
add snippets for 'let' and 'const' statements (ES2015 syntax)

ybiquitous:
partial fix for #134 

For details, see below links:

- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/let
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const


AndreaCrotti:
Cool thanks

ybiquitous:
@AndreaCrotti Thanks for merge! 👍

PR #227, opened 2017-09-27
add 'try...catch' statement snippet to js-mode

ybiquitous:
For details, see below link:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch

### DIFF
--- a/snippets/js-mode/try-catch
+++ b/snippets/js-mode/try-catch
@@ -1,0 +1,11 @@
+# -*- mode: snippet; require-final-newline: nil -*-
+# name: try...catch statement
+# key: try
+# --
+try {
+  $1
+} catch (e) {
+  $2
+}${3: finally {
+  $4
+}}
--- a/snippets/js-mode/try-catch
+++ b/snippets/js-mode/try-catch
@@ -4,7 +4,7 @@
 # --
 try {
   $1
-} catch (e) {
+} catch (err) {
   $2
 }${3: finally {
   $4


AndreaCrotti:
Looks good @ybiquitous 
Only thing I would say is that the exception caught is always called `e`?
Is that a widespread convention?

ybiquitous:
@AndreaCrotti Umm..., I don't have exact statistics, but I see that these are commonly used:

- `err`
  - [Google](https://google.github.io/styleguide/jsguide.html#formatting-nonempty-blocks)
  - [Node.js](https://nodejs.org/api/errors.html#errors_error_propagation_and_interception)
  - [w3schools](https://www.w3schools.com/js/js_errors.asp)
- `e`
  - [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch)
  - [Microsoft](https://docs.microsoft.com/en-us/scripting/javascript/reference/try-dot-dot-dot-catch-dot-dot-dot-finally-statement-javascript)
- others...
  - `error`
  - `ex`

Maybe recently, `err` seems more popular than `e`. But I can use either.

AndreaCrotti:
I would go with `err` and possibly still make it customizable (so something like `${3:err}`, or whatever index you get to).
Thanks!

ybiquitous:
@AndreaCrotti 
I think that `${3:err}` is too customizable, so I will apply `err` (more popular). It's OK?

AndreaCrotti:
Yeah ok let's do `err`, thanks

ybiquitous:
@AndreaCrotti Thanks a lot! 😆

PR #228, opened 2017-09-27
add Promise snippet to js-mode (since ES2015)

ybiquitous:
For details, see below link:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/promise

partial fix for #134


ybiquitous:
@AndreaCrotti Thanks!

PR #230, opened 2017-10-24
Add snippet dir with symbol indirection

### LINKED ISSUE #224: Error when updating the yasnippet package from Melpa
I don't think this is really a bug with this package, but generally when a new version gets installed and you try to open a file you get an error like:

```
File mode specification error: (file-error Opening directory No such file or directory /home/andrea/.emacs.d/elpa/yasnippet-snippets-20170920.1234/snippets/js-mode)
yas--subdirs: Opening directory: No such file or directory, /home/andrea/.emacs.d/elpa/yasnippet-snippets-20170920.1234/snippets/js-mode```
```

After a `yas-reload-all` it's all sorted even though you still see this in *Messages*
```
[yas] Check your `yas-snippet-dirs': /home/andrea/.emacs.d/elpa/yasnippet-snippets-20170920.1234/snippets is not a directory
```

And looking at the `yas-snipppet-dirs` it contains both the hold and the new package paths.
I wonder if there is anything to alleviate this problem, the code I used for the discovery is basically the same used by the other snippet packages (even if I didn't notice the same issue there).
Any idea?

npostavs:
Should fix #224, but not really tested properly yet.
```
* yasnippet-snippets.el (yasnippet-snippets-dir): Define with
defconst, set to the "snippets" directory.
(yasnippet-snippets-initialize): Add `yasnippet-snippets-dir' as a
symbol.  Drop the boundp check, as we are already doing (require
'yasnippet).
```

### DIFF
--- a/yasnippet-snippets.el
+++ b/yasnippet-snippets.el
@@ -31,21 +31,21 @@
 
 (require 'yasnippet)
 
-(setq yasnippet-snippets-dir
-      (file-name-directory
-       ;; Copied from ‘f-this-file’ from f.el.
-       (cond
-        (load-in-progress load-file-name)
-        ((and (boundp 'byte-compile-current-file) byte-compile-current-file)
-         byte-compile-current-file)
-        (:else (buffer-file-name)))))
+(defconst yasnippet-snippets-dir
+  (expand-file-name
+   "snippets"
+   (file-name-directory
+    ;; Copied from ‘f-this-file’ from f.el.
+    (cond
+     (load-in-progress load-file-name)
+     ((and (boundp 'byte-compile-current-file) byte-compile-current-file)
+      byte-compile-current-file)
+     (:else (buffer-file-name))))))
 
 ;;;###autoload
 (defun yasnippet-snippets-initialize ()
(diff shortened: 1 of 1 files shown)


AndreaCrotti:
Ok nice thanks as soon as I manage to give it a try I'll merge it @npostav!

yasnippet-snippets.el
```diff
@@ -31,21 +31,21 @@
...
+     (:else (buffer-file-name))))))
 
 ;;;###autoload
 (defun yasnippet-snippets-initialize ()
-  (let ((snip-dir (expand-file-name "snippets" yasnippet-snippets-dir)))
-    (when (boundp 'yas-snippet-dirs)
-      (add-to-list 'yas-snippet-dirs snip-dir t))
-    (yas-load-directory snip-dir)))
+  (add-to-list 'yas-snippet-dirs 'yasnipet-snippets-dir t)
+  (yas-load-directory yasnipet-snippets-dir))
```
AndreaCrotti: there is a typo here @npostavs but fixing that locally it seems to all work.
If you fix that I'll merge it, thanks

npostavs: oops, fixed.

PR #231, opened 2017-10-24
Move some C/++ specific stuff to its own subdir

npostavs:
See https://github.com/joaotavora/yasnippet/issues/580#issuecomment-338540349


SteveJobzniak:
Thanks! Very useful, but I doubt that `c-lang-mode` will be accepted since a future mode may use that name. I bet @AndreaCrotti will want another name for this reorganizing. Something without `-mode` at all I guess...

AndreaCrotti:
Ah nice @npostav 
so I think it's fine but yes maybe `c-lang-mode` suggests that there should be a `c-lang-mode` which actually doesn't exist.
So maybe something like `c-lang-common` instead? (Or I'm open to other suggestions of course)

SteveJobzniak:
I like `c-lang-common` a lot.

npostavs:
> I like c-lang-common a lot. So +1.

`c-lang-common` it is.

AndreaCrotti:
Great thanks

PR #234, opened 2017-10-28
Add customization group

notetiene:
Hi,
I added a customization group that I think will be useful if someone uses the `.yas-setup.el` facilities.  This would allow setting custom snippet preferences (like name of author or the project copyright holders) for snippet expansion.

The group in itself doesn't do anything, but it helps grouping customizable options.

### DIFF
--- a/yasnippet-snippets.el
+++ b/yasnippet-snippets.el
@@ -48,6 +48,16 @@
       (add-to-list 'yas-snippet-dirs snip-dir t))
     (yas-load-directory snip-dir)))
 
+(defgroup yasnippet-snippets nil
+  "Options for yasnippet setups.
+
+This is useful for customizing options declared in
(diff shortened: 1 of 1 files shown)


AndreaCrotti:
Ah that's a nice idea thanks!
I wonder if it's good to add it empty or we should first have at least one example of customisable variable first, the name of author could be a good example if you'd like to add it?

AndreaCrotti:
I'm still a bit reluctant adding this until we actually have something to customize (otherwise the group will stay empty), but I'll just keep the PR here for reminder and merge it as soon as we have something to customize.

AndreaCrotti:
I solved the conflicts in https://github.com/AndreaCrotti/yasnippet-snippets/pull/301
and I'll merge that one, thanks

PR #235, opened 2017-10-29
Update template

YourFin:
Moved spaces that were part of the default values out of them as they are always going to be left in, and moved the quotes outside of the docstring default as they will usually be there.


AndreaCrotti:
Cool thanks!

PR #237, opened 2017-11-10
Add method and accessor (getter/setter) snippets for ES6

ybiquitous:
For details, see below links:

- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Method_definitions
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/set

partial fix for #134

### DIFF
--- a/snippets/js-mode/getter
+++ b/snippets/js-mode/getter
@@ -1,0 +1,8 @@
+# -*- mode: snippet; require-final-newline: nil -*-
+# name: getter
+# key: get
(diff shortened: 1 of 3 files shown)


AndreaCrotti:
nice thanks

ybiquitous:
Thank you! 😄

PR #239, opened 2017-12-01
Add snippet for elisp package

kchenphy:


snippets/emacs-lisp-mode/header
```diff
@@ -0,0 +1,33 @@
+# -*- mode: snippet; -*-
+#contributor: kchenphy <https://www.github.com/kchenphy>
+#name: package header
+#key: header
+#condition: (= (1 (point))
```
npostavs: I think you mean `(bobp)`

kchenphy: I accidentally pushed an earlier commit.

In the new commit, I removed the condition part altogether.

AndreaCrotti:
Sure might be handy

PR #243, opened 2018-01-01
Added Rust snippets "eprint" and "eprintln"

ludwigpacifici:
Hello Andrea,

This PR adds eprint/eprintln Rust snippets which behave like print/println.

* [eprintln](https://doc.rust-lang.org/std/macro.eprintln.html): Macro for printing to the standard error, with a newline.
* [eprint](https://doc.rust-lang.org/std/macro.eprint.html): Macro for printing to the standard error.


AndreaCrotti:
nice thanks

PR #244, opened 2018-01-02
Add snippets for swift

jojojames:


snippets/swift-mode/uiviewcontrollerlifecycle
```diff
@@ -0,0 +1,25 @@
+# -*- mode: snippet -*-
+# name: uiViewControllerLifecycle
+# key: uiViewControllerLifecycle
+# --
+// MARK: UIViewController lifecycle
```
AndreaCrotti: this one seems quite long, do you always have to write all this code together as a single unit?

snippets/swift-mode/uitableviewdelegate
```diff
@@ -0,0 +1,34 @@
+# -*- mode: snippet -*-
+# name: uiTableViewDelegate
+# key: uitableViewDelegate
+# --
+// MARK: UITableViewDelegate
```
AndreaCrotti: same here, also there is nothing to expand here, is this really just boilerplate code that everyone has to copy & paste around (sorry I know nothing about Swift & co)

AndreaCrotti:
Thanks a lot @jojojames 
Some snippets seem more framework specific other than actual Swift only (forgive my ignorance).
Are they really all generic enough that most Swift developers would use these snippets?

snippets/swift-mode/uiviewcontrollerlifecycle
jojojames: Yeah. It's very common to pull those methods in and ViewControllers are very common in iOS land. (One of the big 3 (M.V.Controllers)

jojojames:
Yup! Most of them take inspiration from https://github.com/burczyk/XcodeSwiftSnippets so there's some precedence in them being useful. :)

AndreaCrotti:
Ok makes sense

PR #246, opened 2018-01-06
Add snippets for kotlin

jojojames:


snippets/kotlin-mode/maino
```diff
@@ -0,0 +1,8 @@
+# -*- mode: snippet -*-
+# name: JvmStatic main function
+# key: maino
+# --
```
AndreaCrotti: uhm not sure it make sense to have a different snippet which is the same as `main` except for the `@JvmStatic` on top.
Is it something that it's so commonly used?
And what does the `o` in `maino` is for anyway?

snippets/kotlin-mode/fun1
```diff
@@ -0,0 +1,7 @@
+# -*- mode: snippet -*-
+# name: Function with one parameter
+# key: fun1
+# --
+fun $1($2: $3): $4 {
```
AndreaCrotti: I would just have a generic fun where you can step into the parameters and declare as many as you want.
`fun $1(${2:parameters}): $3..` or something like that.
Would not add a different snippet for each possible number of arguments (even if I agree that possibly 1-2 arguments are quite common cases, but it's not something done for other snippets in this repository at least)

snippets/kotlin-mode/maino
jojojames: These are from https://github.com/JetBrains/kotlin/blob/master/idea/resources/liveTemplates/Kotlin.xml

I'm not sure how common it is but they were in ^ so I included them.

snippets/kotlin-mode/fun1
jojojames: Same as above^, I can remove them and use your suggestion if you think that's better.

One value add of these multi-parameter snippets is these functions expand to

fun functionName(parameterOne: parameterOneType).

AndreaCrotti: Mm yes I would remove these two fun1 and fun2, and just make fun a bit more generic to just define as many parameters as you like..

snippets/kotlin-mode/maino
AndreaCrotti: well I would remove `maino` as well, doesn't make much sense to me to have this kind of snippet that simply annotates another snippet..

snippets/kotlin-mode/fun1
jojojames: Done.

snippets/kotlin-mode/maino
jojojames: Done.

AndreaCrotti:
Thanks @jojojames , merging

PR #248, opened 2018-01-21
Remove paren-wrapping snippets from emacs-lisp-mode

npostavs:
There are several snippets whose expansion is nothing more than

    (KEY $0)

where KEY is the snippet's #key field.  These snippets do nothing but
save you typing in the parentheses.  Using a package like paredit, or
smartparens would be more suited to this.  There are currently a lot
of snippets in emacs-lisp-mode, which makes browsing through the list
more difficult, so these least useful ones should be removed.

### DIFF
--- a/snippets/emacs-lisp-mode/append
+++ b/snippets/emacs-lisp-mode/append
@@ -1,5 +1,0 @@
-#contributor: Xah Lee (XahLee.org)
-#name: append
-#key: append
-# --
-(append $0)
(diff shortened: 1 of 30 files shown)


npostavs:
PS I think some others like `and` and `or` (keys of `a` and `o`, respectively) might be worth removing too, but one step at a time.

AndreaCrotti:
Uhm yeah I agree it doesn't really add any value, specially when the key is the same as the actual function name, so happy to get rid of them..

PR #250, opened 2018-01-22
Latex: add snippet for description

mcraveiro:
Similarly to itemize and enumerate, the description snippet creates a list of items, with a description associated. Copied from the enumerate snippet.


AndreaCrotti:
Thanks!

PR #251, opened 2018-01-26
Add bibtex-mode snippets

wynro:
Not much else to say. If there is any problems just ask

### DIFF
--- a/snippets/bibtex-mode/article
+++ b/snippets/bibtex-mode/article
@@ -1,0 +1,18 @@
+# -*- mode: snippet -*-
+# name: article
+# key: article
+# expand-env: ((yas-indent-line 0))
+# --
+@article{ ${title},
+  author  = {${author}},
+  title   = {${title}},
+  journal = {${journal}},
+  year    = {${year}}${,
+  volume  = {${volume}}}${,
+  number  = {${number}}}${,
+  pages   = {${pages}}}${,
+  month   = {${month}}}${,
+  note    = {${note}}}
+}
+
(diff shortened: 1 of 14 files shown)


npostavs:
The `((yas-indent-line 0))` setting looks a bit odd to me. I would probably write `nil` rather than `0` to disable. Is `bibtex-mode`'s indent function unhelpful? Perhaps it should be disabled for the mode globally in a `.yas-setup.el` file rather than per-snippet, similar to how it's done for python-mode:

https://github.com/AndreaCrotti/yasnippet-snippets/blob/b42c2b670bdd761b9c1c232998ebf4bbc5b914e8/snippets/python-mode/.yas-setup.el#L38-L39

wynro:
As far as I now, the default `bibtex-mode` doesn't even have indentation. The `<tab>` is binded to a movement function, and the function `indentation-according-to-mode` simply puts everything in the first line. The indentation is more an stylistical choice, and to aid visualization. If I don't put it, yasnippet defaults to indent the snippet, and aligns everything to the first line.

About not putting it in `.yas-setup.el`, I simply didn't knew about the file. Putting `0` instead of the idiomatic `nil` is also related with my newbiness. I'll do both changes and push again. ¿Its better if I merge it with the other commit, or do you prefer them separated?

npostavs:
> ¿It's better if I merge it with the other commit, or do you prefer them separated?

I defer to @AndreaCrotti on that, but I guess it's easier to squash than unsquash.

PR #252, opened 2018-02-12
Can we use JIT snippet loading?

linkenneth:
I found that yasnippet now supports JIT snippet loading. Can we make this the default setting when we initialize snippets? From some very basic benchmarking this reduces my load time of yasnippet-snippet autoloads from ~2200 ms to ~400 ms. Further compiling the snippets pushes this down to ~360 ms.

### DIFF
--- a/yasnippet-snippets.el
+++ b/yasnippet-snippets.el
@@ -45,7 +45,7 @@
 ;;;###autoload
 (defun yasnippet-snippets-initialize ()
   (add-to-list 'yas-snippet-dirs 'yasnippet-snippets-dir t)
-  (yas-load-directory yasnippet-snippets-dir))
+  (yas-load-directory yasnippet-snippets-dir t))
 
 ;;;###autoload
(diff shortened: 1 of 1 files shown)


AndreaCrotti:
Ah yes absolutely, I didn't know about this option.
But I just wonder if I need to do something to support older versions of yasnippet?
Probably not and I guess it has been available for a while now, so I think we can just go with it, any thoughts @linkenneth @npostavs ?

npostavs:
Jit-loading has been around for a while. joaotavora/yasnippet@922304ffdfc3ea87cc6a1fa34e72afb092c66206 reversed the sense of the parameter, but that was before 0.8.0 (which you're already requiring as a minimum) so there's no problem from a versioning perspective.

AndreaCrotti:
Ah that's good to hear, no reason not to merge it for me then, thank you both

linkenneth:
Thanks! 🥇

PR #255, opened 2018-02-22
Add snippets for bazel-mode

Sunhick:
bazel mode snippets for c++


AndreaCrotti:
The only risk with these kind of snippets is that they can easily tend to be outdated and/or not contain all the possible fields that are actually available.

Anyway since it could probably be useful for someone else I think it makes sense to merge this, thanks for your contribution @Sunhick

PR #256, opened 2018-02-23
Fix indentation python-mode/function_docstring_numpy

servitoroftheoutergods:
the indentation was wrong, using only one space instead of 4.

### DIFF
--- a/snippets/python-mode/function_docstring_numpy
+++ b/snippets/python-mode/function_docstring_numpy
@@ -5,7 +5,7 @@
 # group: definitions
 # --
 def ${1:name}($2):
-    \"\"\"$3
-    ${2:$(python-args-to-docstring-numpy)}
-    \"\"\"
-    $0
+ \"\"\"$3
+ ${2:$(python-args-to-docstring-numpy)}
+ \"\"\"
+ $0


npostavs:
But what about the `# NOTE: Use minimum indentation, because Emacs 25+ doesn't dedent docstrings.`? (I'm pretty sure I wrote that note, but I don't recall the exact details right now)

servitoroftheoutergods:
the indentation was not a multiple of four.
the snippet  function_docstring_numpy has the correct indentation.
function_docstring is now with the same indentation.

servitoroftheoutergods:
I use Emacs 25.3.1, and the docstring is working.

npostavs:
> I use Emacs 25.3.1, and the docstring is working.

Did you check with `python-indent-offset` set to `2` (as described in joaotavora/yasnippet#750)?

servitoroftheoutergods:
```

Result when using ‘python-indent-offset’ = 1:

def name():
 """
 
 """
 
Can’t guess python-indent-offset, using defaults: 2


Result when using ‘python-indent-offset’ = 2:

def name():
  """
  
  """
  
Can’t guess python-indent-offset, using defaults: 3


Result when using ‘python-indent-offset’ = 3:

def name():
   """
   
   """
   
Can’t guess python-indent-offset, using defaults: 4


Result when using ‘python-indent-offset’ = 4:

def name():
    """
    
    """
    
Can’t guess python-indent-offset, using defaults: 5


Result when using ‘python-indent-offset’ = 5:

def name():
     """
     
     """
     
Can’t guess python-indent-offset, using defaults: 6


Result when using ‘python-indent-offset’ = 6:

def name():
      """
      
      """
      
Can’t guess python-indent-offset, using defaults: 7


Result when using ‘python-indent-offset’ = 7:

def name():
       """
       
       """
       
Can’t guess python-indent-offset, using defaults: 8


Result when using ‘python-indent-offset’ = 8:

def name():
        """
        
        """

```

servitoroftheoutergods:
:O plot twist, function_docstring_numpy is wrong.

servitoroftheoutergods:
thanks @npostavs  for the patience ^^ .

PR #258, opened 2018-03-19
Rename files that are prohibited in Debian packages

sten0:
Strictly speaking only hyphen and back-quote are prohibited.  I left the underscores alone, but it might be more consistent to rename them too.


AndreaCrotti:
Thanks @sten0 but these snippets don't have a `key` set, so this change would also change how you have to trigger the snippets.
If you just add the key in the changed snippets to avoid breaking how it works now, I'm happy to merge it!

sten0:
I decided to add missing keys for all markdown-mode snippets, because might as well, right :-)  Unfortunately I'm not sure about the h1.?, h2.?, and hr.? ones, so this commit represents my best guess.  Please let me know if it needs more work.  Thanks @AndreaCrotti !

AndreaCrotti:
Well I think it's fine @sten0 

The only thing that happens if you add two snippets with the same key is that yasnippet will ask you which one you want to expand of all the ones that are matching a specific key.

If these snippets are actually kind of related and it makes sense for them to use the same key then I'm happy about that as well.

And lastly we could probably also rename all the files to a more meaningful name now that you added the keys everywhere (`_` and `__` are not great file names for example) but it's not something that has to be done in this PR sure.

sten0:
Hi, @AndreaCrotti 

Good to hear!  I've taken care of the most useful renames.  My gut tells me they should use the same key, but I'm honestly not sure about, for example `h1.1` vs `h1.2`.
```
# ${1:Header 1} #

$0
```
using the same `key: h1` as
```
${1:Header 1}
${1:$(make-string (string-width yas-text) ?\=)}

$0
```
when the descriptions are `Header 1 (#)` vs `Header 1 (=)`

@erickgnavar and @thiagowfx, care to comment?

Cheers,
Nicholas

AndreaCrotti:
Well I think noone really remembers h1.2 for example, so maybe we can just merge them, and possibly I can simplify further if one of the two options is not really useful.
Ok then I'll merge this now in the meanwhile, thanks for the melpa stable badge as well!

PR #260, opened 2018-03-24
Add various snippets for org-mode

neymarsabin:
This will add various snippets for org-mode like title, date, reveal and more.


snippets/org-mode/readme.org
```diff
@@ -0,0 +1,61 @@
...
+* Introduction 
+	orgTemplates is a collection of basic templates used for org-mode of EMACS.
+	
+* Use
+	Using orgTemplates is very simple.Just clone this repository.Then copy all files inside orgTemplates to your emacs configuration folder excluding hidden files.
+#+BEGIN_SRC bash 
+cd orgTemplates
+rsync -av --exclude=".*" . ~/.emacs.d/snippets/org-mode/
+#+END_SRC
+Then,tell emacs to reload all the snippets.For that do *M-x yas-reload-all*
```
AndreaCrotti: this file should be removed or updated at least?

snippets/org-mode/reveal_all
```diff
@@ -0,0 +1,43 @@
+# -*- mode: snippet -*-
+# name: reveal_all
+# key: <rvl
+# --
+#+OPTIONS: reveal_center:t reveal_control:t reveal_height:-1
```
AndreaCrotti: don't think we need to have all these options..

snippets/org-mode/rv_background
```diff
@@ -0,0 +1,8 @@
+# -*- mode: snippet -*-
+# name: reveal_single_colored_background
+# key: <rsb
+# --
+
+		:PROPERTIES:
```
AndreaCrotti: indentation seems wrong here

snippets/org-mode/sourcecode
```diff
@@ -0,0 +1,7 @@
+# -*- mode: snippet -*-
+# name: sourcecode
+# key: <s
```
AndreaCrotti: This one already exists and it's from org-mode, so I think it's not needed

AndreaCrotti:
Some very nice snippets but maybe a bit too much, and the thing is that by default `TAB` is bound to `org-cycle` so I normally haven't really used yasnippet at all in org-mode files.
How do you have it set up @neymarsabin ?

neymarsabin:
Thanks for the comments @AndreaCrotti, Surely **TAB** is bound to org-cycle. But most of the snippets start with `<` which won't be a problem while using with TAB. I use most of them every time I use org-mode and these can be handy most of the time. :+1:

snippets/org-mode/reveal_all
neymarsabin: Most of these options are removed. And these options are available from org-mode export too. I kept them here just in case.

snippets/org-mode/table
```diff
@@ -0,0 +1,8 @@
+# -*- mode: snippet -*-
+# name: table
+# key: <ta
+# --
+#+CAPTION: ${1: caption of the table}
+#+ATTR_HTML: :width 100%
```
AndreaCrotti: attr html is not really always needed imho, I never used it at least

AndreaCrotti:
Looks good, the only snippet I would not merge is the `reveal_all` one possibly, and I left another comment as well.
If you fix these two little things I'm happy to merge it, thanks for your contribution!

AndreaCrotti:
Thanks!

neymarsabin:
Thanks @AndreaCrotti and happy to contribute more.

PR #261, opened 2018-03-27
fix indent function haskell

jadermcs:
Without this the second line gets indented when pressing TAB.

### DIFF
--- a/snippets/haskell-mode/function
+++ b/snippets/haskell-mode/function
@@ -1,6 +1,7 @@
 # -*- mode: snippet -*-
 # name: fun
 # key: fun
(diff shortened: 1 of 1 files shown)


snippets/haskell-mode/function
```diff
@@ -1,6 +1,7 @@
 # -*- mode: snippet -*-
 # name: fun
 # key: fun
+# expand-env: ((yas-indent-line 'fixed))
 # --
 ${1:function-name} :: ${2:type}
-$1 ${3:arguments} $0
```
npostavs: You shouldn't add a newline at the end though, I think.

AndreaCrotti:
Thanks!

PR #262, opened 2018-03-30
rust: add snippets for crate, pub fns, and constructor

balajisivaraman:
Some useful snippets for Rust mode that I've found wanting.

- The `extern crate` and the `new` ones are hopefully straightforward. (These are equivalent to the ones I found useful in [Vim Snippets](https://github.com/honza/vim-snippets/blob/master/snippets/rust.snippets).)
- The `pub` variants of all the function snippets is something I've wanted for a long time. Right now I sometimes forget to add `pub` in front and have to go back to the beginning line. Other times, I'll usually add `pub` and then begin the snippet. With the newly added snippets, I hope to provide a canonical way to do them that are analogous to the non-`pub` counterparts.

### DIFF
--- a/snippets/rust-mode/crate
+++ b/snippets/rust-mode/crate
@@ -1,0 +1,5 @@
+# -*- mode: snippet -*-
+# name: extern crate
+# key: ec
+# --
+extern crate ${0:name};
--- a/snippets/rust-mode/new
+++ b/snippets/rust-mode/new
@@ -1,0 +1,7 @@
+# -*- mode: snippet -*-
+# name: fn main() { ... }
+# key: new
+# --
+pub fn new($1) -> ${2:Name} {
+   $2 { ${3} }
+}
--- a/snippets/rust-mode/pfn
+++ b/snippets/rust-mode/pfn
@@ -1,0 +1,7 @@
+# -*- mode: snippet -*-
+# name: pub fn name() { ... }
+# key: pfn
+# --
+pub fn ${1:name}($2) {
+    $0
+}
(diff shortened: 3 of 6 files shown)


AndreaCrotti:
It's a bit of a shame that these snippets are exactly the same as existing snippet declarations with just a prefixed `pub`.
Apparently it's possible to have "subsnippets" but the stacked expansion is kind of disabled by default
```
yas-triggers-in-field is a variable defined in ‘yasnippet.el’.
Its value is nil

Documentation:
If non-nil, allow stacked expansions (snippets inside snippets).
```

So maybe it's not worth trying to be too smart.
Another option would be to always add `pub` in front allowing to remove it easily if not needed, which can be done just by modifying the existing snippets like

`${1:pub }fn ....`

@balajisivaraman would that be ok? How much is the split roughly between public and not public functions?

balajisivaraman:
@AndreaCrotti, Thank you for having a look at this. Sorry for not responding earlier, but I'm currently out and unable to address your concerns immediately. I should be back in a week. I will have a look at what is the split between `pub` and non`-pub` functions in some Rust codebases I work with and update the PR accordingly.

balajisivaraman:
@AndreaCrotti, I've checked in a sample of the major repos I regularly use (`riggrep`, `rust-lang`, `crater` etc.), and the lean seems to be more towards non-public functions. I would still prefer to have the `pub` variants, since the Vim snippets module also has them. However, if you feel strongly about this, I wouldn't mind removing them and updating the PR to only have the `extern crate` snippet for the time being.

AndreaCrotti:
Yes ok I think it's fine, sounds sufficiently mnemonic to me as well to prefix with a p.

balajisivaraman:
@AndreaCrotti, Thank you so much! :+1:

PR #265, opened 2018-04-07
php-mode: add support for class attributes

rrooij:
This makes it easier to generate getters and setters for class
attributes in PHP.

### DIFF
--- a/snippets/php-mode/get
+++ b/snippets/php-mode/get
@@ -1,0 +1,10 @@
+# -*- mode: snippet -*-
+# contributor: Robin de Rooij <rderooij685@gmail.com>
+# name: get
+# key: get
+# group: definitions
+# --
+public function get${1:$(capitalize yas-text)}()
+{
+    return \$this->$1;
+}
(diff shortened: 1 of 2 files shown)


AndreaCrotti:
I'm not a PHP expert but from a quick search it doesn't look like it's such common practice to implement it this way.
I saw some examples of `__get` and `__set` but never just `get` and `set`, any resources to point out this is kind of standard?

The other thing I don't like too much is that the order becomes
- var declaration
- getter
- setter

While maybe you might want a different layout like adding all the variables first.
Thanks for your contribution anyway of course, but unless it's something quite generic that more or less everyone could use I'd rather not merge this one..

AndreaCrotti:
Ok I see thanks for the extra information.
The simplest solution is just to add two separate snippets for the getter and the setter, which could be triggered by `set` and `get` for example?
In that way you would not really have any issues with the layout anymore, and allow people to use these snippets more freely.

rrooij:
That is a good solution indeed. Should I rebase in this pull request and squash it with the commit?

AndreaCrotti:
Yes ok looks good thanks

PR #267, opened 2018-04-10
Add snippets for Racket

codingbum:
A few snippets for the [Racket programming language](https://racket-lang.org/).

### DIFF
--- a/snippets/racket-mode/do
+++ b/snippets/racket-mode/do
@@ -4,4 +4,4 @@
 # --
 (do ([${1:name} ${2:init} ${3:step}]$4)
     (${5:stop-predicate} ${6:finish})
-  ${7:body})$0
+  $0)
--- a/snippets/racket-mode/do
+++ b/snippets/racket-mode/do
@@ -1,0 +1,7 @@
+# -*- mode: snippet -*-
+# name: (do ([... ... ...]...) (... ...) ...)
+# key: do
+# --
+(do ([${1:name} ${2:init} ${3:step}]$4)
+    (${5:stop-predicate} ${6:finish})
+  ${7:body})$0
--- a/snippets/racket-mode/for
+++ b/snippets/racket-mode/for
@@ -1,0 +1,5 @@
+# -*- mode: snippet -*-
+# name: (for... (...) ...)
+# key: for
+# --
+(for$1 (${2:for-clause}) ${3:body})$0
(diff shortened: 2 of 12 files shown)


AndreaCrotti:
Looks good but I'm not super convinced by having snippets that take the `body` as well.
I mean suppose you write quite a long `cond` with a lot options, you would have to write everything  still while expanding the snippet.
I would rather simply have $0 be the body, so at that point you write the more bulky part you expanded the whole snippet already.

What do you think @codingbum ?

snippets/racket-mode/for
```diff
@@ -0,0 +1,5 @@
+# -*- mode: snippet -*-
+# name: (for... (...) ...)
+# key: for
+# --
+(for$1 (${2:for-clause}) $0)
```
AndreaCrotti: why is the $1 just next to the for here?
Does it mean you might want to add something to "for"?

AndreaCrotti:
Well don't see why not, what's the problem with the $0 there in the do loop @codingbum ?
One slight issue for me is that some of these snippets have way too much stuff to expand in general, for example the do loop just mentioned has 6 things to expand.

I don't think there is a "rule" for this but normally most snippets involve maximum 3-4 tabs..
It would be nice to stay around that, even if it means the snippet is a bit less precise, but the important thing after all is to give a good skeleton not to help you write everything imho.

snippets/racket-mode/for
codingbum: It's because racket has a multitude of for/... 
```
(for/list ...) ;; build a list
(for/vector ...) ;; build a vector
(for/sum ...) ;; sum
(for/and ...) ;; checks if all iterations return (not #f)
...
```
https://docs.racket-lang.org/reference/for.html?q=for#%28part._.Iteration_and_.Comprehension_.Forms%29

snippets/racket-mode/do
```diff
@@ -0,0 +1,7 @@
+# -*- mode: snippet -*-
+# name: (do ([... ... ...]...) (... ...) ...)
+# key: do
+# --
+(do ([${1:name} ${2:init} ${3:step}]$4)
```
AndreaCrotti: mm yeah I think that 7 expansions is way too much, I would just do something like

```(do ([$1] $2)
         ($3)
     $0)```

(with the appropriate placeholder names)
Would that make sense?

AndreaCrotti:
Ok I tried the snippets again and I think it's fine, we can keep also the `do` as it is..
Thanks!

PR #271, opened 2018-04-18
php-mode: fix case in getters and setters

rrooij:
Fix the case of the getter and setter templates. The previous
method "capitalize" turned upper case characters to lower case
elsewhere in the string.


AndreaCrotti:
Great thanks @rrooij

PR #275, opened 2018-05-19
Don't expand Go main package definitions at beginning of buffer

juergenhoetzel:
Go programs typically start with a main package definition:
```Go
package main // Don't expand

func main() {
}
```

### DIFF
--- a/snippets/go-mode/main
+++ b/snippets/go-mode/main
@@ -1,5 +1,6 @@
 # -*- mode: snippet -*-
 # name: func main()
+# condition: (progn (forward-line 0) (not (and (eq (point-min) (point)) (looking-at-p "package"))))
 # key: main
 # --
 func main() {


AndreaCrotti:
That could be fine for me, but you mean you are writing `package main` yourself and then pressing `tab` and you don't want it to expand?

juergenhoetzel:
> That could be fine for me, but you mean you are writing package main yourself and then pressing tab and you don't want it to expand?

Exactly. 👍

AndreaCrotti:
Yes ok but just to understand, why do you tab there considering it's at the end of the line, and you don't need to indent anyway?

juergenhoetzel:
> Yes ok but just to understand, why do you tab there considering it's at the end of the line, and you don't need to indent anyway?

I use `company-yasnippet` to automatically complete snippets. 
Also I mechanically  press <key>TAB</key> when company  completes to check if there is a useful snippet. This predicate would prevent me from doing this (in this case invalid) completion.

AndreaCrotti:
Yes ok sounds reasonable, thanks!

PR #276, opened 2018-06-01
Add method snippet to go-mode

Kunde21:
Used to add a method to a go struct with documentation comment.

### DIFF
--- a/snippets/go-mode/method
+++ b/snippets/go-mode/method
@@ -1,0 +1,9 @@
+# -*- mode: snippet -*-
+# name: method
+# key: mthd
+# contributor: @Kunde21
+# --
+// $2 ${5:...}
+func (${1:recv}) ${2:name}(${3:args}) $4 {
+  $0
+}


AndreaCrotti:
It doesn't seem very different from the `func` snippet, why do we need a different one?
And from my understanding Go is no that Object Oriented so `method` would not be the right name anyway, methods are normally related to OO style classes

Kunde21:
It's not very different, the difference being the method receiver definition before the function name.  When creating a struct method, the receiver is the first element and usually the first thing on the dev's mind -- cumbersome to add after the `func` snippet is used.  I've been using this snippet locally for some time, but finally got around to submitting it upstream.

Note:  Go is object oriented in that it has [struct methods](https://tour.golang.org/methods/1), interfaces, and embedding; what it doesn't have is inheritance.

AndreaCrotti:
Ah cool ok then sounds good to me, not too sure mthd is very mnemonic but I think that's good enough

PR #277, opened 2018-06-17
Snippets for dart-mode

l7ssha:
Hello,
I've added few snippets for dart-mode. For now there no dart-mode snippets.


snippets/dart-mode/for
```diff
@@ -0,0 +1,8 @@
+# -*- mode: snippet -*-
+# name: for
+# key: for
+# expand-env: ((yas-indent-line 'fixed) (yas-wrap-around-region nil))
+# --
+for(var i = ${1:0}; i ${2:< 10}; i${3:++}) {
+  ${4:Body}
```
AndreaCrotti: could we not put $0 here instead of at the end of the for?

snippets/dart-mode/fori
```diff
@@ -0,0 +1,8 @@
+# -*- mode: snippet -*-
+# name: fori
+# key: fori
+# expand-env: ((yas-indent-line 'fixed) (yas-wrap-around-region nil))
+# --
+for(var ${1:obj} in ${2:collection}) {
```
AndreaCrotti: same thing here

snippets/dart-mode/func
```diff
@@ -0,0 +1,7 @@
+# -*- mode: snippet -*-
+# name: func
+# key: func
```
AndreaCrotti: maybe `fun` or even just `f` is better?

snippets/dart-mode/funca
```diff
@@ -0,0 +1,7 @@
+# -*- mode: snippet -*-
+# name: funca
+# key: funca
```
AndreaCrotti: considering it's an async function, maybe `afun` / `af` would seem more mnemonic to me, what do you think?

snippets/dart-mode/getset
```diff
@@ -0,0 +1,7 @@
+# -*- mode: snippet -*-
+# name: getset
+# key: getset
+# --
+${1:Type} _${2:Name};
```
AndreaCrotti: is that such a common pattern to have it this way?

snippets/dart-mode/getter
```diff
@@ -0,0 +1,6 @@
+# -*- mode: snippet -*-
+# name: getter
+# key: get
+# --
```
AndreaCrotti: same question here

AndreaCrotti:
Great thanks @l7ssha left a few comments and questions

l7ssha:
I've applied some of your notes @AndreaCrotti . Thanks for review.

snippets/dart-mode/getset
l7ssha: Yes, getters are commonly used.  This is generic pattern but it applies pretty well in this case.

snippets/dart-mode/getter
l7ssha: Same here

snippets/dart-mode/getset
l7ssha: And if you mean '_Name' notation - '_' in Dart means 'private' in C like programming languages, so it has to be like that.

AndreaCrotti:
Great thanks

PR #279, opened 2018-06-19
Add elisp snippet using uppercase

rememberYou:
By default, `org-mode` inserts blocks in uppercase. I allowed myself to modify the snippet to suit this standard.


AndreaCrotti:
Nice, thanks

PR #284, opened 2018-07-14
c++-mode: fix `tfm` snippet ignored by changing duplicate descriptions

Hi-Angel:
There were multiple snippets going under the name `transform`, and because of that the `tfm` one, which was actually the `transform` function, didn't work. Fix that by changing names in duplicates to what they actually are.


AndreaCrotti:
Thanks!

PR #286, opened 2018-08-17
Update rust-mode/ifl

mssdvd:


snippets/rust-mode/ifl
```diff
@@ -2,4 +2,6 @@
 # name: if let pattern = expression { ... };
 # key: ifl
 # --
-if let ${1:pattern} = ${2:expression} { $0 };
\ No newline at end of file
+if let ${1:pattern} = ${2:expression} {
```
AndreaCrotti: fine for me if that's the more standard way of writing this

mssdvd: https://github.com/rust-lang-nursery/fmt-rfcs/blob/master/guide/expressions.md

PR #292, opened 2018-10-03
Use the same c++ class name placeholder in all c++-mode snippets

mineo:
The snippets previously used Class, MyClass, Name and Type. Now it's just Name
in all of them, based on the class11 snippet.

Also replaces `yas/substr` calls in c++-mode/class with `yas-c++-class-name`.

### DIFF
--- a/snippets/c++-mode/class
+++ b/snippets/c++-mode/class
@@ -5,7 +5,7 @@
 class ${1:Name}
 {
 public:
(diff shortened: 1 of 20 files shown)


AndreaCrotti:
Looks good thanks a lot!

AndreaCrotti:
And sorry for the delay

PR #294, opened 2018-10-09
Few C snippets

Wathu:
My first commit to "real" repo ever, so maybe isn't good enough.

### DIFF
--- a/snippets/c-mode/puts
+++ b/snippets/c-mode/puts
@@ -1,0 +1,5 @@
+# -*- mode: snippet -*-
+# name: puts
+# key: puts
+# --
+puts("${1:string}");
--- a/snippets/c-mode/puts
+++ b/snippets/c-mode/puts
@@ -2,4 +2,4 @@
 # name: puts
 # key: puts
 # --
-puts("${1:string}");
+puts(${1:string});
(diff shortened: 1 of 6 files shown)


snippets/c-mode/puts
```diff
@@ -0,0 +1,5 @@
+# -*- mode: snippet -*-
+# name: puts
+# key: puts
+# --
+puts(${1:string});
```
AndreaCrotti: I'm not a great fan of snippets with only one thing to expand, in this case it would only really save one keystroke (the `(`), and maybe the ";" at the end.

snippets/c-mode/strlen
```diff
@@ -0,0 +1,5 @@
+# -*- mode: snippet -*-
+# name: strlen
+# key: strlen
+# --
+strlen(${1:string});
```
AndreaCrotti: same thing here, unless we maybe shorten it (but not sure how it would be shortened) I would not include such trivial snippets

AndreaCrotti:
A few extra comments, FYI we also have `cc-mode` which is a parent of c-mode, and a few others like:
```$ ack "cc-mode" .
snippets/c-mode/.yas-parents
1:cc-mode

snippets/c++-mode/.yas-parents
1:cc-mode

snippets/csharp-mode/.yas-parents
1:cc-mode
```

These snippets seem C specific though so maybe it's fine this way.

Wathu:
I'll remember about single arg snippets ;-)
Thank You for extra comments.

I'am newbie with C and Emacs, so I will be adding c snippets gradually, if they are useful.

PR #300, opened 2019-01-07
Modify markdown-mode snippet 'plus' to remove new line

Ryota0312:
For example, if you use bullet snippet 3 times, useless 2 new brank lines are inserted as follows.
```markdown
+ text1
+ text2
+ text3
+
(blank line)
(blank line)
```
I think it is more convenient not to insert a newlines as follows.
```markdown
+ text1
+ text2
+ text3
+
```


AndreaCrotti:
Sure makes sense

PR #302, opened 2019-02-07
Update if, ife in sh-mode

sei40kr:


snippets/sh-mode/if
```diff
@@ -2,7 +2,7 @@
 # name: if
 # key: if
 # --
-if ${1:[ -f file]}
-   then ${2:do}
+if ${1:[ -f file]}; then
```
AndreaCrotti: First of all thank you for the contribution @sei40kr !

I prefer the current version of both snippets tbh, is there any style guide or reason in particular to change it to the one you suggest?

AndreaCrotti:
I'll close this in the meanwhile, feel free to re-open if you feel strongly about this

PR #304, opened 2019-02-27
stdlib: remove newline at end of file

nocriw:
Other include snippets as `unistd`, `string`, `stdio`, ... don't have a newline at the end of the file and I expect this to be the correct behavior. On the other side, `stdlib` has a newline at the end of file.

This fixes the inconsistency in the `stdlib` snippet. :)


AndreaCrotti:
Thanks

PR #305, opened 2019-02-28
Refresh JS snippets

wyuenho:
This PR 

- Add React/Redux/Jest/GraphQL snippets for rjsx-mode (ported from [vscode-es7-javascript-react-snippets](https://github.com/dsznajder/vscode-es7-javascript-react-snippets/blob/master/snippets/snippets.json))
- Refresh ES5/6 snippets, prefer ES2015+ (ported from [vscode-es7-javascript-react-snippets](https://github.com/dsznajder/vscode-es7-javascript-react-snippets/blob/master/snippets/snippets.json))
- Remove some old framework specific JS snippets and duplicates


snippets/js-mode/anfn
```diff
@@ -0,0 +1,10 @@
+# -*- mode: snippet -*-
+# uuid: 289f0557-4e11-47c5-bcbf-1105bbec41ce
```
AndreaCrotti: Ah cool, I didn't even know there was `uuid`.
But well we only have 1 so far in the whole project and even in this PR is not really on every snippet.

I think it makes more sense if they are always there, which could be done by a big batch generation maybe, but we can maybe remove them for now from this PR.

snippets/rjsx-mode/React-Native/rnc
```diff
@@ -0,0 +1,19 @@
+# -*- mode: snippet -*-
+# uuid: c4aa3b3e-82af-434b-acc5-b08bf018e0aa
+# contributor: Jimmy Yuen Ho Wong <wyuenho@gmail.com>
+# name: reactNativeComponent
+# key: rnc
```
AndreaCrotti: can we add a `group: react` to all the react related snippets please?

AndreaCrotti: ah right I noticed now that they are in `rjxs-mode`, but JSX does not imply necessarily React right?

wyuenho: JSX implies React 95+% of the time. `rjsx-mode` also implies you are probably writing React code.

snippets/js-mode/anfn
wyuenho: UUID is as good as globally unique now and forever, I don't think it matters when you add it to a snippet. I think it exists only to distinguish snippets for different modes having the same prefix or name.

snippets/rjsx-mode/React-Native/rnc
wyuenho: It wouldn't make sense to group React-Native snippets under React. React snippets is valid when programming under React and React-Native, but React-Native snippet is only valid under React-Native. It also wouldn't make sense to see React-Native snippets under the React group when programming for the Web. Ideally, there should be some mechanism to hide the React-Native group altogether when the file is detected to be under a Web project, and merge the React-Native group under React when the file is under a mobile project, but this is probably best left to the user to customize or there will be a butch of glue code that suits one but not the other.

wyuenho:
@AndreaCrotti How does this look?

AndreaCrotti:
Ok @wyuenho LGTM, thanks a lot for the contribution

PR #306, opened 2019-03-01
add snippets for google provider in terraform

mwasilew2:
Signed-off-by: Michal Wasilewski <mwasilewski@gmx.com>

### DIFF
--- a/snippets/terraform-mode/google/app_engine_resources/google_app_engine_application
+++ b/snippets/terraform-mode/google/app_engine_resources/google_app_engine_application
@@ -1,6 +1,6 @@
 # -*- mode: snippet -*-
 # name: google_app_engine_application
-# key: google_app_engine_application
+# key: goog_app_engine_application
 # --
 
 resource "google_app_engine_application" "${1:name}" {
--- a/snippets/terraform-mode/google/app_engine_resources/google_app_engine_application
+++ b/snippets/terraform-mode/google/app_engine_resources/google_app_engine_application
@@ -1,0 +1,10 @@
+# -*- mode: snippet -*-
+# name: google_app_engine_application
+# key: google_app_engine_application
+# --
+
+resource "google_app_engine_application" "${1:name}" {
+  project     = "${2:project_id}"
+  location_id = "${3:location_id}"
+}
+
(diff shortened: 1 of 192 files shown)


AndreaCrotti:
Very nice @mwasilew2 
Just a couple of things, I'm not sure super sure about the keys, which seem very long to type in, and maybe we should add a `group:` directive in all these snippets (just "google" as group would be enough).

mwasilew2:
@AndreaCrotti   I agree, the keys are quite long, I'm hesitant myself. I actually considered shortening them and tried a few approaches, e.g. `gc_instance` for `google_compute_instance` , but it quickly became convoluted, e.g. what about `google_compute_instance_data`? `gc_instance_data` or `gcd_instance`. It also required knowing the system used for shortening names.

On the other hand, if these snippets are used with autocompletion, you actually don't have to type that much and long keys become an advantage as they get rid of any ambiguity (you know exactly what resource will be created). In addition, snippets become "self discoverable" - you don't have to be aware of the fact that emacs has any of them for terraform. As soon as you start typing a name of a resource, emacs will suggest using a snippet for it. If the keys and names of resources were not an exact match, you would have to know which resource you're after, if snippets for it are available and how to use them.

I think that ultimately it's a UX question, so opinions will vary and each approach has pros and cons ;)

as for groups, I added `.yas-make-groups`, let me know if you'd rather have `:groups` inside files, I'm happy to do that

AndreaCrotti:
`.yas-make-groups` looks good actually, but could you do the same for the existing AWS snippets?
otherwise now it's inconsistent.

For the names well I would maybe even `goog_` for google, and `aws_` for aws, which could be enough @mwasilew2

mwasilew2:
I can't seem to find any AWS snippets, in terraform directory I see only generic terraform snippets, nothing specific to AWS. could you point me to them?

changed keys from `google_` to `goog_`

@AndreaCrotti back to you

AndreaCrotti:
Ah that's true @mwasilew2 I assumed they were aws specific for some reason but they are actually generic.
Ok then I think we can merge this, quite a lot of useful snippets for whoever might be using Google cloud.

PR #311, opened 2019-04-13
Elixir: make hcall, hcast, and hinfo snippets more general

huynh91:
Hi, 
The current version of handle_call and handle_cast is too specific. Most of the time, I have to change variable name (state) and delete code.


AndreaCrotti:
Makes sense, thanks!

PR #316, opened 2019-05-03
add equation and align (multiline equation environment)

dakling:
I added two snippets for latex that are probably useful for anyone editing math in latex.


AndreaCrotti:
Cool thanks

Issue #303, opened 2019-02-12
org mode prefers lowercase keywords now


memeplex:
Since some months ago the org mode codebase is preferring lowercase blocks and keywords. See for example [this commit](https://code.orgmode.org/bzg/org-mode/commit/13424336a6f30c50952d291e7a82906c1210daf0). Also, the template expansion based on 9.2 hardcodes `#+begin_/#+end_` in lowercase. Unfortunately, the snippets in this collection are all uppercase. Since there surely are and will be people that prefer the uppercase convention, what can be done to provide both options? Offer two sets of snippets?  In case of providing just one, it's going to be time to move to lowercase.

memeplex:
See also http://lists.gnu.org/archive/html/emacs-orgmode/2019-02/msg00146.html

equwal:
PERSONALLY I AM ALL FOR THE LOWERCASE CONVENTION, SINCE UPPERCASE CONVENTIONS ARE FALLING OUT OF STYLE EVERYWHERE.

[CONSIDER THIS ISSUE DEALT WITH, I THINK](https://github.com/AndreaCrotti/yasnippet-snippets/pull/322).

AndreaCrotti:
This is now sorted

PR #322, opened 2019-06-23
Downcase org snippets

equwal:
https://github.com/AndreaCrotti/yasnippet-snippets/issues/303

Also, I removed the comment from the "video" snippet, since it probably should not be there.

PERSONALLY I AM ALL FOR THE LOWERCASE CONVENTION, SINCE UPPERCASE CONVENTIONS ARE FALLING OUT OF STYLE EVERYWHERE.

I also downcased the names, since there isn't really any reason to maintain, for example; HTML over html.

### DIFF
--- a/snippets/org-mode/author
+++ b/snippets/org-mode/author
@@ -2,4 +2,4 @@
 # name: author
 # key: <au
 # --
-#+AUTHOR: $0
+#+author: $0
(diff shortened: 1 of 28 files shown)


AndreaCrotti:
Nice thanks

PR #328, opened 2019-07-15
Adds several flutter snippets for dart-mode

bradyt:
Some initial progress on adding snippets for flutter. For example, the following sample from https://api.flutter.dev/flutter/widgets/StatefulWidget-class.html would be easier to input.

``` dart
class YellowBird extends StatefulWidget {
  const YellowBird({ Key key }) : super(key: key);

  @override
  _YellowBirdState createState() => _YellowBirdState();
}

class _YellowBirdState extends State<YellowBird> {
  @override
  Widget build(BuildContext context) {
    return Container(color: const Color(0xFFFFE306));
  }
}
```

### DIFF
--- a/snippets/dart-mode/didchangedependencies
+++ b/snippets/dart-mode/didchangedependencies
@@ -1,0 +1,10 @@
+# -*- mode: snippet -*-
+# name: didChangeDependencies
+# key: dcd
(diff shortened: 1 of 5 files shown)


AndreaCrotti:
Thanks

PR #330, opened 2019-07-23
Add perl use snippet.

equwal:
This heading is standard in perl code, I actually have it auto-insert (using the autoinsert for Emacs) whenever I open a .pl file. A lot of perl programs have no file ending, since they are meant as executables, so I'd like to have it in my yasnippets.


AndreaCrotti:
Yes why not can be useful, thanks

PR #333, opened 2019-08-15
rust: make fns/pfns like the rest of the function templates (fix #323)

### LINKED ISSUE #323: rust: pfns / fns don't create a block
fn / fnr snippets create a function block but pfns / fns end with ;. It makes more sense that creates a block.

r-darwish:


AndreaCrotti:
Thanks!

PR #336, opened 2019-08-31
Org not underline

equwal:
org-subscript

This kind of underlining now gets turned into a subscript in org, so I removed it (since the raw code in the old one is `external_link`.


AndreaCrotti:
Thanks!

PR #337, opened 2019-09-03
c++: make cin recognize whether 'using namespace std' exists

hankail05:


AndreaCrotti:
Thanks @hankail05 , very nice solution

PR #339, opened 2019-09-08
Source block without any language pre-defined (like the elisp snippet)

equwal:
I have deleted the ` emacs-lisp :tangle yes` portion of the `elisp` snippet enough times to make we want to have this pull request. It adds a new `src` snippet.

Would it be better to also remove the `elisp` snippet, since it is a bit arbitrarily specific? Org mode can support any language you want in the src blocks, so I think this is a good thing.

### DIFF
--- a/snippets/org-mode/src
+++ b/snippets/org-mode/src
@@ -1,0 +1,6 @@
+# -*- mode: snippet -*-
+# name: src
+# key: src_
+# --
+#+begin_src $0
+#+end_src


AndreaCrotti:
There used to be a snippet like that, but you can get the same result with just

`<s TAB`

Which it's the standard org-mode way I would say, so not sure it's worth adding that snippet

npostavs:
FYI, I believe newer org modes (e.g., the one currently in Emacs master at least) have switched to a new way (so `<s TAB` won't work).

https://www.reddit.com/r/emacs/comments/ad68zk/get_easytemplates_back_in_orgmode_92/

equwal:
> FYI, I believe newer org modes (e.g., the one currently in Emacs master at least) have switched to a new way (so `<s TAB` won't work).
> 
> https://www.reddit.com/r/emacs/comments/ad68zk/get_easytemplates_back_in_orgmode_92/

I have a recent version of org and `<s TAB` does not expand.

equwal:
The new command is `C-c C-,` (as noted in the [reddit comment](https://www.reddit.com/r/emacs/comments/ad68zk/get_easytemplates_back_in_orgmode_92/ede34jv?utm_source=share&utm_medium=web2x)), which I was not aware of. I will start using that.

I am closing this request. Thank you.

PR #340, opened 2019-09-08
Move .nosearch to ./snippets subdirectory

benzwick:
Otherwise the command normal-top-level-add-subdirs-to-load-path
cannot be used to add yasnippet-snippets to the load-path.
This way, the package yasnippet-snippets is made available
but the actual snippets are not loaded as lisp code
(which would cause an error).

See also: 45a652e ("Create .nosearch", 2014-10-04)


AndreaCrotti:
Thanks @benzwick , but I don't quite understand why you need to use that command to add these snippets?
You don't use `yas-snippet-dirs` to add more snippets directories?
Or I misunderstood something?

benzwick:
I added `yasnippet-snippets` as a git submodule and load it like this:
```
;; Add Git submodules to load path
(let ((default-directory
	(concat user-emacs-directory (convert-standard-filename "submodules/"))))
  (normal-top-level-add-subdirs-to-load-path))

(require 'yasnippet)
(require 'yasnippet-snippets)

(add-hook 'prog-mode-hook #'yas-minor-mode)
```
If `.nosearch` is in the top level directory (as it is now) the `yasnippet-snippets` package cannot be loaded using the commands `normal-top-level-add-subdirs-to-load-path` and `(require 'yasnippet-snippets)` because Emacs does not search the `yasnippet-snippets` subdirectory. Without `.nosearch` Emacs tries to load the snippets which causes an error. Moving `.nosearch` to the `snippets` subdirectory solves the problem by allowing Emacs to search the top level `yasnippet-snippets` directory (to find `yasnippet-snippets.el`) but not the `snippets` subdirectory.

AndreaCrotti:
Ah I see, I haven't used git submodules for Emacs packages in a long time I forgot about it.

Anyway makes sense I can merge it, thanks

PR #341, opened 2019-09-14
Delete obviously erroneous (and often annoying) trailing whitespace.

equwal:
I only removed whitespace that was not either in some kind of a block, or behind the cursor. I just did this and then manually checked each file's matching lines:

```shell
snippets $ grep -R "[ ]$"
```


AndreaCrotti:
Sorry @equwal now it conflicts with your other PR which I just merged, would you mind rebasing?
 I'll merge it as soon as you've done it, thanks!

equwal:
It is a bit wonky, but I think it is okay now.

PR #342, opened 2019-09-14
Renamed "switch" (which is not in lisp) to "cond" which is.

equwal:
Not sure why this was named `switch`, it is a `cond`. `cond` is not close to switch in ALGOL derivates at all.


AndreaCrotti:
Thanks

PR #343, opened 2019-09-20
Py docstring type annotation

WJCFerguson:
A few tweaks to the Python snippet fn for handling of args to generating docstrings, e.g. the `function_docstring` (key `fd`) snippet.

Before this it was upset by newlines in arg lists, e.g.
``` python
def foo(
        arg1,
        arg2,
):
```

And in addition, it didn't know what to do with type annotations, e.g:
``` python
def foo(
        arg1: int,
        arg2: str = 'default string',
):
```

It now excludes the type annotation from the docstring.

(I could see an argument for removing the default handling - my personal preference is to avoid duplicating the information - I'd be happy to modify the PR to do that, but who knows which users use and like this.)

### DIFF
--- a/snippets/python-mode/.yas-setup.el
+++ b/snippets/python-mode/.yas-setup.el
@@ -2,9 +2,12 @@
 (defvar yas-text)
 
 (defun python-split-args (arg-string)
-  "Split a python argument string into ((name, default)..) tuples"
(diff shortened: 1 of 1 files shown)


AndreaCrotti:
Thanks, I'll test it soon but otherwise looks good to me

WJCFerguson:
Closing - been around way too long and I don't have the motivation to drive it..

PR #346, opened 2019-10-05
added a crystal (crystal-lang) snippets for crystal-mode

yilkalargaw:
These are snippets which I use when I program with crystal. Since the language is very similar to ruby most of them are modified from the ruby snippets.

### DIFF
--- a/snippets/crystal-mode/any
+++ b/snippets/crystal-mode/any
@@ -1,0 +1,5 @@
+# -*- mode: snippet -*-
+# name: any? { |...| ... }
+# group : collections
+# --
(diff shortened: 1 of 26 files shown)


AndreaCrotti:
Great @yilkalargaw 
Would you mind adding an explicit `key` to add the snippets?
I don't like too much to depend on the filename for that, but other than that looks good

yilkalargaw:
Yhea you're right specific key tags are better than file names. I have added the keys for these crystal snippets.

AndreaCrotti:
Great thanks

PR #351, opened 2019-11-04
Change variable order in Python try snippets

r-darwish:
By changing the code inside the try block to $0 it is possible to wrap the action region with a
try/except block when yas-wrap-around-region is set to t.


AndreaCrotti:
Good idea, thanks!

PR #352, opened 2019-11-07
Don't add anonymous functions to hooks, use named functions instead

stepnem:
This makes the functions transparent (clear where they came from) and
accessible (compilable, removable, reusable, redefinable).


AndreaCrotti:
That's a nice improvement thanks!

PR #353, opened 2019-11-10
Add new snippets for sh-mode: case, select

sei40kr:
# Description

Add new snippets for `sh-mode`:
- `case`
- `select`


AndreaCrotti:
Thanks!

PR #355, opened 2019-11-14
Fix subscripting link

equwal:
This script makes an annoying subscript.

To test, open an org file and type `something_littlething`


equwal:
Oops, I already sent a pull req for this.

PR #356, opened 2019-11-18
Add a generic org src block

r-darwish:


AndreaCrotti:
Should we not use the new mechanism `C-c C-,` now for this purpose @r-darwish ?

r-darwish:
I personally don't like the idea of org inventing its own snippets mechanism. I use yasnippets in almost every mode so for me it makes sense to keep using it even in org mode.

AndreaCrotti:
Yeah ok makes sense, just not sure about the key, in another PR it was added with "<src" which maybe is better since it's consistent to how org used to work, what do you think @r-darwish ?

r-darwish:
I'm okay with `<src`

AndreaCrotti:
ok I'll close this then and I'll merge the other proposal

PR #357, opened 2019-11-26
Add snippets for Dockerfile, Fish shell, JavaScript, Perl, and Org

sei40kr:
## Dockerfile

- `alpine-dockerize`

## Fish shell

- `bang`
- `block`
- `bp`
- `for`
- `function`
- `if`
- `ife`
- `sw`
- `while`

## JavaScript

- `flow`

## Perl

- `bang`
- `enc`
- Remove the shebang from `use` and rename it `strict`

## Org

- `src`

### DIFF
--- a/snippets/org-mode/src
+++ b/snippets/org-mode/src
@@ -1,0 +1,7 @@
+# -*- mode: snippet -*-
+# name: src
+# key: <src
+# --
+#+BEGIN_SRC $1
+  $0
+#+END_SRC
--- a/snippets/perl-mode/strict
+++ b/snippets/perl-mode/strict
@@ -1,9 +1,8 @@
 # -*- mode: snippet -*-
-# name: use
-# key: use_
+# name: strict
+# key: strict
 # contributor: Spenser Truex
 # --
-#!/usr/bin/perl
 use warnings;
 use strict;
 $0
--- a/snippets/dockerfile-mode/alpine-dockerize
+++ b/snippets/dockerfile-mode/alpine-dockerize
@@ -1,0 +1,12 @@
+# -*- mode: snippet -*-
+# name: dockerize for Alpine Images
+# key: alpine-dockerize
+# expand-env: ((yas-indent-line 'fixed) (yas-wrap-around-region nil))
+# --
+ENV DOCKERIZE_VERSION ${1:v0.6.1}
+
+RUN apk update && \
+  apk --no-cache add openssl && \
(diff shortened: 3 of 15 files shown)


snippets/perl-mode/strict
```diff
@@ -0,0 +1,7 @@
+# -*- mode: snippet -*-
+# name: use strict; use warnings
+# key: strict
```
AndreaCrotti: there is already a snippet like that called `use` at the moment.
We could merge them or maybe rename the other one?

snippets/org-mode/src
```diff
@@ -0,0 +1,7 @@
+# -*- mode: snippet -*-
+# name: src
+# key: <src
+# --
+#+BEGIN_SRC $1
```
AndreaCrotti: This was also added in another PR, I think I'll add it separately

snippets/dockerfile-mode/dockerize
```diff
@@ -0,0 +1,11 @@
+# -*- mode: snippet -*-
+# name: dockerize
+# key: dockerize
+# --
+ENV DOCKERIZE_VERSION v0.6.1
+
+RUN apk update && \
```
AndreaCrotti: Didn't know about this dockerize,but it looks like it supports many OS, this is just the Alpine specific version.
At least I would make the name and key more specific.

AndreaCrotti:
Thanks a lot @sei40kr I left a few comments!

AndreaCrotti:
@sei40kr if you could address the few comments I left I'd be happy to merge this PR

sei40kr:
@AndreaCrotti I updated the PR. Please review :pray:

AndreaCrotti:
I'm still not convinced about the `dockerize` snippet, at least it should be clear that it's just the alpine version, can we rename it to reflect that @sei40kr ?

sei40kr:
@AndreaCrotti 
Screenshot from 2019-12-10 23-27-24
I used `yas-choose-text` so that we can select Linux or Alpine Linux when expanding the snippet.

AndreaCrotti:
I think that's not complete because the `linux` (which is ubuntu) doesn't use apk but uses apt.

I kind of prefer if they were two different snippets, or just leave one snippet for alpine for now.

sei40kr:
@AndreaCrotti 
Sorry I've just updated the PR. I renamed it `alpine-dockerize`.

AndreaCrotti:
Nice thanks!

PR #358, opened 2019-12-11
fix initializer

peterhoeg:
Currently, the symbol yasnippet-snippets-dir ends up being added to the
yas-snippet-dirs list instead of the value of yasnippet-snippets-dir.


AndreaCrotti:
Argh that's a silly mistake, and it has been there for a while!
Thanks a lot!

PR #359, opened 2019-12-25
Improvements and fixes for React & Redux snippets

sei40kr:
## Description

- Remove legacy React snippets:
  - SFC without arrow function
  - Unsafe React component methods
- Fix `TM_FILENAME_BASE` in some snippets


AndreaCrotti:
I don't know much about react/redux but I'll do my best to review.
One question though, these legacy snippets you are removing, can't they still be useful to someone using an older version of react maybe?

Or it's unlikely that anyone would still need them?

AndreaCrotti:
I had another look and it looks good to me, thanks @sei40kr

PR #360, opened 2019-12-25
Improve dockerize snippet

sei40kr:
## Description

- Enable to choose which OS to use, Ubuntu or Alpine, in dockerize snippet.


AndreaCrotti:
Nice, thanks!

PR #363, opened 2020-01-25
Added a lot more snippets for terraform

CsBigDataHub:


AndreaCrotti:
thanks a lot @CsBigDataHub !

However I'm a bit scared of this PR, it's the biggest I've ever seen probably not sure how to tackle it.

Out of curiosity, are the terraform scripts auto generated from something?
Would be good to see how they were generated in that case.

And well there is more than just terraform, could we at least split different things in different PRs by any chance?

CsBigDataHub:
@AndreaCrotti , Terraform snippets are are generated from this source - https://github.com/juliosueiras/vim-terraform-snippets/tree/master/terraform using this ruby gem https://github.com/cartolari/snipmate_to_yas

Yes we can split the PR for terrform and GO. I can create different branches from the commit history and create new pull requests. How does that sound?

AndreaCrotti:
Yes @CsBigDataHub splitting it in 3 PRs would be great, would be much easier to deal with.

I'm also a bit "worried" about how long it actually takes to load so many snippets, I guess anyone not using terraform would not notice at all, but I never had so many snippets for one mode so I'm not sure exactly how fast it is.

CsBigDataHub:
@AndreaCrotti
I am using straight to load the package from my fork of the repo and I do not see any delay when I enter terraform mode. I also have lsp mode hook for terraform mode and it's takes a second or two to start and I see that all the snippets loaded already.

CsBigDataHub:
closing this in favor of #379  and #380

shuntakeuch1:
@CsBigDataHub 
That's a good PR.
I was looking for a snippet on AWS.
Is it okay to use the aws snippet from this PR on my local PC?

PR #367, opened 2020-02-16
Add go init snippet

phikal:
Not much to say: This snippet is pretty much the same as main, just without the condition.


AndreaCrotti:
Thanks!

PR #369, opened 2020-03-13
Adds Elixir with snippet

Auralcat:
This is an addition to Elixir's syntax as of version 1.2.0 which wasn't yet included in the current snippet list. 

For any Elixir users over here, what do you think?

### DIFF
--- a/snippets/elixir-mode/wt
+++ b/snippets/elixir-mode/wt
@@ -1,0 +1,8 @@
+# -*- mode: snippet -*-
+# contributor: Miriam Retka (https://github.com/Auralcat)
+# name: wt
+# key: wt
+# --
+with ${1:{:ok, tuple} <- function(arg)}, do
+  $0
+end
--- a/snippets/elixir-mode/wte
+++ b/snippets/elixir-mode/wte
@@ -1,0 +1,10 @@
+# -*- mode: snippet -*-
+# contributor: Miriam Retka (https://github.com/Auralcat)
+# name: wte
+# key: wte
+# --
+with ${1:{:ok, tuple} <- function(arg)}, do
+  $2
+else
+  $0
+end


AndreaCrotti:
Thanks @Auralcat , I think it's a good idea to add it, but this current snippet is imho too prescriptive.
Generally speaking 2-3-4 should imho be the max number of tabs the snippet should expand to, and also for example would you always take two assignments?

Maybe something like (without knowing enough Elixir)

```
with ${1:assignments}
   do
      $2
    else
      $3
  end
end```

So just try to make it a bit simpler and generic and I'm happy to merge it.
(The else branch also would be always there?)

**Comment by Auralcat** [comment]
Got it. For this particular snippet, we need the `else` clause because `with` is a guard statement in Elixir. It processes the initial calls in the upper part and does something with the result in $2. If things go wrong, it runs the `else` clause.

I tried to make the `$1 <- $2` part more generic so far, but I'm having a hard time writing Elisp to transform the text inserted in `${1:assigments}` to something like this:

Input string:
```
some_arg some_function(arg_one, arg_two); another_arg another_function(arg_three, arg_four) 
```

Desired output in the first field:
```
{:ok, some_arg} <- some_function(arg_one, arg_two),
{:ok, another_arg} <- another_function(arg_three, arg_four),
```
I tried using `replace-regexp-in-string` inside the snippet, but for some reason the expression is not changing `yas-text` when expanding the snippet.

**Comment by Auralcat** [comment]
I've included the requested changes in https://github.com/AndreaCrotti/yasnippet-snippets/commit/77bc30cd61a985b31c9ef44340a693734b2f8d40, what do you think? 😄 

Turns out the string transformation got too complicated to do with Elisp and without a mirror field, so I dropped the idea and left a placeholder field in `$1`. I also split the previous snippet in two, one with the `else` clause and one without it, as done in [the VSCode version of the snippets.](https://github.com/florinpatrascu/vscode-elixir-snippets/blob/fb1cfe782e4390c4ed18a09ad8e5d267d43d0466/snippets/snippets.json#L410)

File: snippets/elixir-mode/wt [hunk 1]
```diff
@@ -0,0 +1,8 @@
+# -*- mode: snippet -*-
+# contributor: Miriam Retka (https://github.com/Auralcat)
+# name: wt
+# key: wt
+# --
+with ${1:{:ok, tuple} <- function(arg)}, do
+  $2
```
AndreaCrotti, line 7: the final state is normally just $0

snippets/elixir-mode/wte
```diff
@@ -0,0 +1,10 @@
+# -*- mode: snippet -*-
+# contributor: Miriam Retka (https://github.com/Auralcat)
+# name: wte
+# key: wte
+# --
+with ${1:{:ok, tuple} <- function(arg)}, do
+  $2
+else
+  $3
```
AndreaCrotti, line 9: also here could be $0

AndreaCrotti:
A couple of minor changes @Auralcat and then I'm happy to merge it

snippets/elixir-mode/wte
Auralcat: Done in https://github.com/AndreaCrotti/yasnippet-snippets/commit/2838cdf27241c85524bff9677e808fdb7963992b

snippets/elixir-mode/wt
Auralcat: Done in https://github.com/AndreaCrotti/yasnippet-snippets/commit/2838cdf27241c85524bff9677e808fdb7963992b

AndreaCrotti:
Nice thanks

PR #371, opened 2020-03-30
Add more bibtex snippets

equwal:
All of these snippets are defined in Section 2.1.1 "Regular Types" of the
[biblatex documentation](http://ctan.math.utah.edu/ctan/tex-archive/macros/latex/contrib/biblatex/doc/biblatex.pdf).

I'm also considering the following edits, I'd like to hear your opinion on including them:
- No space between the first `{` and the keyword (currently labeled `${title}`.
- Label that keyword `${key}` instead, since it is a keyword and not a title.

Thanks.

### DIFF
--- a/snippets/bibtex-mode/bookinbook
+++ b/snippets/bibtex-mode/bookinbook
@@ -1,0 +1,22 @@
+# -*- mode: snippet -*-
+# name: bookinbook
+# key: bookinbook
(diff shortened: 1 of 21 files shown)


AndreaCrotti:
Thanks, and fine for the other changes you suggested @equwal

PR #372, opened 2020-04-02
Vhdl mode

serveln:
I included the following snippets for the VHDL-mode:

- architecture
- assignation
- case
- component
- constant
- downto
- entity
- if / if else
- lib
- port
- process
- signal
- to
- type
- when


AndreaCrotti:
Nice one thanks!

PR #374, opened 2020-04-10
Remove inappropriate exec bit for snippets/vhdl-mode/*

sten0:
Hi,

When packaging 0.21 for Debian I noticed that snippets/vhdl-mode/* inappropriately have the executable bit set.  Please merge this fix and tag a new release at your earliest convenience.

Thanks,
Nicholas


AndreaCrotti:
Ah sorry about that @sten0 and thanks for reporting it.
Just out of curiosity, is that a security risk even if none of these things is a script that could run?

Maybe I can add something in CI to check that the permissions are all correct to avoid this happening in the future, since I would not normally notice file permissions from the github PR.

Any other automated check that might be useful?

sten0:
Thank you @AndreaCrotti!

Well, my primary motivation is that it's somewhat of a blocker for uploading to Debian (publicly displays a big warning of shame to anyone to bothers to check my QA pages) rather than infosec paranoia, and I'm not sure if Emacs is vulnerable to stuff like null-byte injection...it is a network-connected program for many (most?) users though...

I like your CI idea to check for +x :-)  Sorry, I can't think of any other checks, other than maybe something like codespell, but Debian tooling has never detected typos or obvious grammar errors in yasnippet-snippets, so I'm not convinced a codespell thing would be worth the effort.

PR #376, opened 2020-04-15
Added \limits to bigcap and bigcup for latex-mode

WhatDothLife:
I've added the possibility to include \limits as a common alternative formatting option.

### DIFF
--- a/snippets/latex-mode/bigcap
+++ b/snippets/latex-mode/bigcap
@@ -2,4 +2,4 @@
 # name: bigcap
 # key: bigcap
 # --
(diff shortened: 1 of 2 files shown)


AndreaCrotti:
@WhatDothLife just for my understanding, since I could not find it easily looking it up, what does bigcap with and without limits look like?
A screenshot of the rendered thing is more than enough..
Thanks!

WhatDothLife:
bigcup
The one on the left is formatted with `\limits`.

AndreaCrotti:
Cool thanks

PR #378, opened 2020-04-27
csharp-mode: add more snippets, clean, and fixes existents ones.

jangoleninsta:
Hi,

CSharp-mode is lacking essential snippets found in Visual Studio and Rider.

Additionally, a few existent snippets required some cleaning and at times fixing its logic.

Some snippets were absolutely opinionated so I converted those to what most .NET devs expect.

For those that need a reference,  Microsoft has an article about Visual Studio snippets:

https://docs.microsoft.com/en-us/visualstudio/ide/visual-csharp-code-snippets?view=vs-2019

PS: I did revise all of them looking for errors four times, but who knows? :D 

Thanks!


snippets/csharp-mode/classFull
```diff
@@ -0,0 +1,21 @@
+# -*- mode: snippet -*-
+# name: Class: ... Name { ctor dtor }
+# key: classfull
+# --
+
+${1:public} class ${2:Name}
+{
```
AndreaCrotti, line 7: I think in this cases both short and long snippets should share the same key, to allow the user to select between them?
The problem is that otherwise you have to remember a million different snippet keys

snippets/csharp-mode/exception
```diff
@@ -1,6 +1,7 @@
 # -*- mode: snippet -*-
 # contributor : Alejandro Espinoza Esparza <aespinoza@structum.com.mx>
 # name: /// <exception cref="..."> ... </exception>
-# key: comment
```
AndreaCrotti, line 4: all the comment snippets are gone now?

snippets/csharp-mode/propertyFull
```diff
@@ -0,0 +1,12 @@
+# -*- mode: snippet -*-
+# name: Property: _var; Acessor Type Name { get => _var; set => _var = value; }
+# key: propfull
```
AndreaCrotti, line 3: also these 3 might share the same key if they would be more used as frequently.

snippets/csharp-mode/exception
lxbarbosa: summary, and its parameters now have a explicit naming instead of the hideous comment{1,2}:

https://github.com/AndreaCrotti/yasnippet-snippets/pull/378/files#diff-6d638a73daf131fa76e992100cbcb48a

PR #379, opened 2020-04-27
[1] Feature/terraform snippets

CsBigDataHub:
@AndreaCrotti 

This is the first breakdown of multiple commits I have made when the previous pull request was active.

This PR contains bunch of terraform snippets


snippets/terraform-mode/acme/.yas-parents
```diff
@@ -0,0 +1 @@
+_-mode
```
AndreaCrotti, line 1: What is this parent mode?

AndreaCrotti:
Nice @CsBigDataHub , just one last question, how do we maintain such a massive amount of snippets?
I mean if you generated them do you need to rerun the generation process periodically to update in case there are new properties for some of the snippets?
Or they are supposed to never change anyway?

snippets/terraform-mode/acme/.yas-parents
CsBigDataHub: > Nice @CsBigDataHub , just one last question, how do we maintain such a massive amount of snippets?
> I mean if you generated them do you need to rerun the generation process periodically to update in case there are new properties for some of the snippets?
> Or they are supposed to never change anyway?

I am using Terraform with Azure for more than year now. I can tell you with confidence that these snippets would not change except for one or two variable that gets added or removed depending on Azure evolution. It would be same for other platforms.

I would defiantly maintain Azure part of terraform because I use it for work. I would expect if people working on other platforms who are using yasnippets would create pull request.

CsBigDataHub: > What is this parent mode?

Every thing here is terraform-mode. I guess I missed to delete this file. Does it effect anything?

staticaland:
This would be pretty awesome. Did you use a script to generate the snippets? It would be very interesting to see how you managed to create all these snippets.

CsBigDataHub:
@staticaland yes. https://github.com/cartolari/snipmate_to_yas

staticaland:
Thanks. It may be possible to generate snippets with the data from `terraform providers schema -json` (https://www.terraform.io/docs/commands/providers/schema.html).

I'll give it a shot.

staticaland:
I just added AWS, Azure and Google snippets to https://github.com/staticaland/yasnippet-terraform by generating them with https://github.com/staticaland/terraform-generate-snippets

Hopefully it will solve the problem of keeping the snippets updated automatically.

I will work on organizing them a little better and might add support for other snippet systems as well.

CsBigDataHub:
@staticaland thanks. Good work on the python scripts.

@AndreaCrotti does it make sense to create a travis ci pipeline to run @staticaland's script periodically to reflect any changes from terraform snippets?

staticaland:
@CsBigDataHub I just ran through almost all the providers, and ended up with a 7GB providers folder and around 6000 snippets. It may be better to keep all this building stuff outside this repo I think. CI pipeline would take very long without some form of caching, and it may use a lot of processing time. I wish Hashicorp had this data available through an API. It would make it much easier. Terraform provider argument reference API.

Hmm, that got me thinking, maybe I should save the JSON for every provider and just commit everything to git.

PR #381, opened 2020-05-08
ENH: Make method scope choosable.

quazgar:
In Java, the method snippet now lets one choose the scope:
- public
- private
- protected
- ""

Also fixed a typo.
* snippets/java-mode/method: Method scope is choosable now.


AndreaCrotti:
Sure makes sense

PR #382, opened 2020-05-08
ENH: New snippet to add the `mode' local variable.

quazgar:
Expanding `-*-' adds a line which sets mode:conf.

* snippets/fundamental-mode/mode-line: Expanding `-*-' adds a line which sets mode:conf.

### DIFF
--- a/snippets/fundamental-mode/mode-line
+++ b/snippets/fundamental-mode/mode-line
@@ -1,0 +1,5 @@
+# name: Mode line
+# key: -*-
+# contributor : Daniel Hornung <d.hornung@indiscale.com>
+# --
+# -*- mode:${1:conf}; -*-


snippets/fundamental-mode/mode-line
```diff
@@ -0,0 +1,5 @@
+# name: Mode line
+# key: -*-
+# contributor : Daniel Hornung <d.hornung@indiscale.com>
+# --
+# -*- mode:${1:conf}; -*-
```
AndreaCrotti, line 5: why conf as default? I guess it doesn't matter too much but something more generic like even `mode` would be better imho.

quazgar: Oh, that was just my personal experience, that for about 95% of all cases where I put the mode line, `conf-mode` is the right thing to do. Probably because for a lot of config-like files, emacs does not recognize them (no .conf suffix, no special rule like /etc/fstab). Examples are apt's sources.list files and nearly everything else in `/etc`.

quazgar: Actually I am in favour of taking the most commonly used mode as default (if more people vote on what it should be), whatever that turns out to be. A generic meta variable like `mode` would help nobody, I think.

AndreaCrotti: Yeah ok why not

PR #384, opened 2020-05-21
Fixed cout template token

wailo:
The first and the second token of the cout template are merged, which is annoying when inserting cout and then typing, the second token `<< \n\` will be removed. This fix separates the second token from the first, so after inserting cout template, the user will be typing directly into the body of the cout.


AndreaCrotti:
Looks good

PR #385, opened 2020-06-13
Drop dependency of "s" library

phikal:
Hi,

I was cleaning up my package and dependency list, when I noticed that in my case I had the "s" library installed just because of this package. When I looked into where it was being used, I noticed it seems like the only place where was used was in a Redux snippet introduced in f52190f3e4d3b8359cfae7ebb6b6328f47fbfbd5. This patch should keep the functionality, while dropping the dependency.


AndreaCrotti:
Nice thanks @phikal

PR #389, opened 2020-07-23
cc-mode/if: escape backslashes

leungbk:
Escaping lets the user choose between writing an unbraced `if` by
pressing `C-d`, or continue with a braced `if` by pressing TAB.
Without this patch, pressing `C-d` leaves an unbalanced right brace.


AndreaCrotti:
Cool thanks

PR #390, opened 2020-07-29
Fix leftover curly brace during expansion in various print snippets

nikgr95:
Right curly brace is leftover upon typing at first tabstop of print,
println, eprint and eprintln.

This commit will ensure that the right curly brace will also be
replaced upon typing at the tabstop.


AndreaCrotti:
Nice thanks

PR #392, opened 2020-09-01
current-date in fundamental-mode

kmt-lnh:
Inspired by [Karl Voit's](https://github.com/novoid) [blog
post](https://karl-voit.at/2016/12/18/org-depend/). It simply inserts
the current date in YYYY-MM-DD bound to 'dd'

Found no conflict with other bindigs.


AndreaCrotti:
Cool thanks

PR #397, opened 2020-10-21
Fix console.log (formatted version)

shevchuk:
Fixes the behavior of a placeholder for formatted console.log.

Steps to reproduce:
1. hit `clo` and expand the snippet in a JS file.
2. start typing the object to inspect

Actual:
only the second` console.log` argument is changed

Expected:
first argument (a label) should match the object name.


AndreaCrotti:
Looks good, thanks!

PR #403, opened 2020-11-15
Add plantuml-mode

Davoodeh:
Sorry, I didn't read any contribution guide and this is my first request here...
I'm also a casual Plantuml user so feel free to comment suggestions.


AndreaCrotti:
Would be nice to set the `key` in every snippet and not depend on the filename, so the filename can be longer and more explicit (and won't break the snippet if you rename the file).
Otherwise looks good to me thanks @Davoodeh

wkok:
I like this addition of plantuml snippets, thanks @Davoodeh  ! Being new to yasnippet in general, I'm just wondering why some snippets have a rather cryptic key, that does not always feel natural/obvious.

For example, why:

- `clr` instead of `color`
- `sp` instaed of `skinparam` , I get that's maybe bit long so ` skinp`
- `uml` instead of `startuml`

None of these are show stoppers in my opinion and I'm sure one will get used to the keys the more you use it, I was just wondering what the reason behind some of the choices was, and if the suggestion of @AndreaCrotti might solve some of these odd ones in that if the key is defined explicitly in the snippet, maybe some keys might be easier to adapt to feel more natural.

PR #409, opened 2020-12-11
js-mode: Add new error constructor

Amorymeltzer:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/Error

----

Personally, I'd use `throw new Error` more often, but I thought this might be more broadly useful.

I've not updated the big html file, since it seems that hasn't been edited in a while?

### DIFF
--- a/snippets/js-mode/error
+++ b/snippets/js-mode/error
@@ -1,0 +1,5 @@
+# -*- mode: snippet; require-final-newline: nil -*-
+# name: new error
+# key: err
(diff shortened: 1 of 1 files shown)


AndreaCrotti:
Ah yes I forgot about the big HTML file, and well that's only generated automatically anyway.
If I automate that process and publish it automatically then we don't have to see it at all in the repo actually.

PR #414, opened 2021-02-07
Add clojure-mode expansion for a Rich comment

dpassen:
- The trailing comma keeps paredit/parinfer from collapsing lines


AndreaCrotti:
Don't think this adds enough tbh since it's just `(comment` pretty much, and you'd need to delete the `,` every time as well.

PR #415, opened 2021-02-08
Update header snippet in emacs-lisp-mode

dalugm:
According to the license text on [gnu], I found this snippet had wrong
punctuation and missed some words.

I think it's better to keep in line with the official website. At least, we
should use `https` instead of `http` for safety.

[gnu]: https://www.gnu.org/licenses/gpl-3.0.html


AndreaCrotti:
Thanks

PR #418, opened 2021-02-16
Nix updates

AndersonTorres:
Nix is now following new, clearer guildelines.


AndersonTorres:
Also, I have indented snippets.html file.

AndreaCrotti:
Nice thanks

PR #420, opened 2021-03-15
Avoid literal 0 in expansions

Amorymeltzer:
There's one more, in ruby-mode's `dow`:

```ruby
downto(${0}) { |${n}|
	$0
}
```

But I don't know what it should be.


snippets/js-mode/pge
```diff
@@ -6,5 +6,5 @@
 # --
 
 get ${1:propertyName}() {
-  return this.${0}
+  return this.${2:name}
```
Amorymeltzer, line 9: Honestly, I prefer `propertyName` again, but figured it shouldn't be repeated?

AndreaCrotti: mm if it's supposed to be the same thing you can just use $1 again directly?
If not also just repeating propertyName is fine I don't see the problem.

Amorymeltzer: @AndreaCrotti It probably depends on how one has constructed their setup.  It wouldn't necessarily be weird to have different capitalization or a more desirable facing term.  I've gone and repeated `propertyName`, which should hopefully fit the middle-ground.

AndreaCrotti:
Nice thanks

PR #425, opened 2021-07-24
snippets/go-mode: add more useful template for test-snippet

yuansl:
Hi,  this commit add a more useful template for test-snippet in go-mode


AndreaCrotti:
As mentioned this is probably not generic enough to be included..

PR #426, opened 2021-08-07
fix: escape bracket inside snippet

Viglioni:
### The problem

This snippet has `{ }` at the constructor step but is not escaped:

#### before escaping:
Screen Shot 2021-08-07 at 17 24 54

#### after escaping:

Screen Shot 2021-08-07 at 17 25 52


AndreaCrotti:
Thanks

PR #429, opened 2021-08-17
remove duplicated scripts

### LINKED ISSUE #399: in c++-mode some snippets don't work because they have a name conflict
Hello. Thanks for Yasnippet!
I noticed that in c++-mode, the snippets with keys "upr" and "tfm" have the same name "transform", thus one of them isn't expanded by my Emacs ("upr"); the same is true for the snippets with keys "vec" and "stv" which have the same name "vector".
This duplication is visible in this list http://andreacrotti.github.io/yasnippet-snippets/snippets.html .
Duplicate names inside a same mode are said to cause problems (here https://github.com/joaotavora/yasnippet/issues/548#issuecomment-68213893 and following).
Changing the name to avoid duplication, recompiling and reloading snippets fixes it in my case.
I'm using a yasnippet package downloaded from MELPA some time in September.

AndreaCrotti:
Fixes #399


AndreaCrotti:
Fixes #399

PR #431, opened 2021-10-02
dockerfile-mode: add 'apt-get'

pataquets:
As per Dockerfile best practices.

### DIFF
--- a/snippets/dockerfile-mode/apt-get
+++ b/snippets/dockerfile-mode/apt-get
@@ -1,0 +1,12 @@
+# -*- mode: snippet -*-
+# name: apt-get
+# key: apt-get
+# --
+ARG DEBIAN_FRONTEND=noninteractive
+RUN \
+  apt-get update && \
+  apt-get install -y \
+    $1 \
+  && \
+  apt-get clean && \
+  rm -rf /var/lib/apt/lists/


pataquets:
@AndreaCrotti ping

AndreaCrotti:
To be honest it's not really generic enough to be included.
I mean it might be useful but it's ubuntu/debian only and probably even there you would not install things always this way.

pataquets:
Thanks for reviewing, @AndreaCrotti.

Sure, it applies only to Debian/Ubuntu, which is what I mainly use. But I've aptly named it `apt-get` (pun intended, sorry but I couldn't help it :smile:). If somebody using other OS wants to add their own, they can just name it `yum-install`, for instance, so they will not conflict with other OSes.

As for the snippet per se, installing packages is quite common and this snippet encompasses a few good practices (such as [deleting apt indexes when done](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#apt-get)) which are quite common. I've sent lots of Dockerfile PRs (both for new files and for improving existing) and never found any issue with them. On the contrary, most of them were missing them (such as or setting `DEBIAN_FRONTEND` env var).

At least, this is from what I could understand from your comments, but I'm not 100% confident I'm getting your point and I might be missing something.

pataquets:
> Thanks for reviewing, @AndreaCrotti.
> 
> Sure, it applies only to Debian/Ubuntu, which is what I mainly use. But I've aptly named it `apt-get` (pun intended, sorry but I couldn't help it 😄). If somebody using other OS wants to add their own, they can just name it `yum-install`, for instance, so they will not conflict with other OSes.
> 
> As for the snippet per se, installing packages is quite common and this snippet encompasses a few good practices (such as [deleting apt indexes when done](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#apt-get)) which are quite common. I've sent lots of Dockerfile PRs (both for new files and for improving existing) and never found any issue with them. On the contrary, most of them were missing them (such as setting `DEBIAN_FRONTEND` env var).
> 
> At least, this is from what I could understand from your comments, but I'm not 100% confident I'm getting your point and I might be missing something.

@AndreaCrotti Any feedback on this? I have it added in my private snippets collection, but I still think it might be helpful for others.

PR #432, opened 2021-10-03
dockerfile-mode: add 'dockerfile-image-name' local file variable

pataquets:
As per Readme for dockerfile-mode.
Notice I'm not using obsolete `docker-image-name`, but `dockerfile-image-name`. See spotify/dockerfile-mode#68.


AndreaCrotti:
The `key` is quite long @pataquets , I  think to make it a snippet it should save some typing, but atm it doesn't make much difference typing the whole thing out, otherwise I think it's fine.

PR #434, opened 2021-10-04
php-mode: add match keyword

takeokunn:
add match keyword in php-mode snippet
https://www.php.net/manual/ja/control-structures.match.php


AndreaCrotti:
Thanks!

takeokunn:
Thank you!

PR #436, opened 2021-11-19
Add graphviz snippets

OlavStornes:
While Company supports graphviz in linting, a big portion of my workflow was easier with snippets instead. Hopefully they are useful for others as well.

[Documentation](https://graphviz.org/)


AndreaCrotti:
Not too sure about this tbh, if there is already completion with company I would not repeat all the same possible choices in yasnippet snippets, and these snippets seem a bit too complex.
Unless someone else thinks it's a good idea I would maybe not merge this PR sorry.

OlavStornes:
Hi, thanks for answering.

I can understand that issue and can understand that they introduce some bloat.

If i remove everything company can do (everything except graph creation) and strip those down to a bare minimum with much less complexity, could that be better?

OlavStornes:
Thank you for your time.

I've refined it down in my last commit (8a6985d480bfc724b47cb00105ad3a0f5e9acfc8) to be more of a complement to company, and rather focus on delivering the initial scaffolding needed.

PR #437, opened 2022-01-05
Add snippets for defclass, defgeneric, and defmethod

contrapunctus-1:
This is my first time using as well as writing snippets, hopefully it's all in order :sweat_smile:


contrapunctus-1:
I made this when I didn't know that there was a`common-lisp-snippets` package. Welp.

PR #441, opened 2022-02-17
modified ifmain to create the main function with possible renaming

jakob1379:
For quicker setup if main files.


AndreaCrotti:
Mm sorry I don't really agree in a way, the current snippet just add these two lines, which is useful in itself.
Adding another function if you need to is trivial and probably would not be useful in most cases where you already have  some code, so I'll close this.

PR #443, opened 2022-03-25
Update snippet dir after modifying it

jcs090218:
It seems like without this line; you would always have to call `yas-reload-all` in your configuration? This should lower your Emacs' start up time if you have a tone of snippets to load.


AndreaCrotti:
Yes you are probably right, it's been an issue forever tbh but wasn't bad enough to look into detail.
Yeah I think your solution would be better, thanks

snowman:
This makes `(yas-load-directory yasnippet-snippets-dir t)` redundant

PR #447, opened 2022-07-03
add racket-mode snippets: module, module+, module*, define-syntax-rule and syntax-rules

xgqt:
Refs:
- https://docs.racket-lang.org/reference/module.html
- https://docs.racket-lang.org/reference/stx-patterns.html


xgqt:
fixed module snippets keys

Issue #387, opened 2020-06-17
commentblock in emacs-lisp mode is strange


quazgar:
## Observed behaviour ##

For me, using the `cob` (`commentblock`, from the prog-mode section) snippet in elisp mode produces this result:

```
;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
                                        ;             Some comment            ;
;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
```

## What seems wrong to me ##

There are lots of whitespaces before the first `;` and it's only a single `;`, not as I expected `;;`.  Opposed to this behaviour, `comment-region` prepends double semicolons.

## What the snippet should do ##

- All three lines should start with `;;` at the beginning of the line.

## Discussion ##

If the current behaviour is as intended, that's fine for me and the issue could be closed.

quazgar:
On further investigation, this seems to be a general problem with yasnippet in lisp-mode: If the return value of a `${1:$( ... )}` expression starts with a single semicolon, that line will be indented to column 40, which makes no sense for comment blocks.

So as a workaround I propose to put a double semicolon in front. I will post a pull request soon.

quazgar:
See PR #454

PR #454, opened 2022-12-11
FIX: Single semicolon at beginning was indented.

quazgar:
The workaround here is to add a double semicolon, iff the start comment is a single semicolon.  This also required some fixing of the whitespace calculations.  This commit is for issue
387 (https://github.com/AndreaCrotti/yasnippet-snippets/issues/387).


AndreaCrotti:
Ah nice thanks it annoyed me for a while as well, I'll try it out

PR #457, opened 2023-01-15
Fixes try-src-prefix definition is void

jaybarra:
Fixes Clojure `ns` snippet.

Original behavior yields
```clojure
(ns Symbol’s function definition is void: try-src-prefix)
```
Found and fixed in:
GNU Emacs 28.2 (build 1, x86_64-apple-darwin18.7.0, NS appkit-1671.60 Version 10.14.6 (Build 18G95)) of 2022-09-12


AndreaCrotti:
Nice thanks

PR #468, opened 2023-04-11
*yasnippet-snippets.el: avoid duplicate loading on initialize the snippets

sunlin7:
Hi,

The `(yasnippet-snippets-initialize)` will be called twice for the `yasnippet-snippets` installed by `package.el`.
It happened for the `eval-after-load` is marked as `;;;autoload`, it will appear in the file `yasnippet-snippets-autoloads.el` which is generated by the `package,el`, then the function will be called in the autoloads file and the main file.
https://github.com/AndreaCrotti/yasnippet-snippets/blob/8e4c521252501dd9ad71ea78fae14683ab7a14cb/yasnippet-snippets.el#L73-L75

This change will detect the duplicate call and ignore it.
Please help review and merge the the patch. Thanks.

### DIFF
--- a/yasnippet-snippets.el
+++ b/yasnippet-snippets.el
@@ -49,8 +49,9 @@
   ;; NOTE: we add the symbol `yasnippet-snippets-dir' rather than its
   ;; value, so that yasnippet will automatically find the directory
   ;; after this package is updated (i.e., moves directory).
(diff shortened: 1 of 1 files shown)


AndreaCrotti:
Thanks!

PR #473, opened 2023-06-21
very pedantic spelling correction

mrhappy200:
I believe a more correct way of phrasing this is "link to the video/image" instead of "link of the video/image". Sorry for the (possibly incorrect) correction. I just noticed it while browsing snippets and it sounded wrong.


mrhappy200:
Yay, my first PR :)

PR #475, opened 2023-07-22
Updates to Python Snippets

theoparashkevov:
- Class General
- Class General with Documentation
- Bubble Sort
- Insertion Sort
- Merge Sort
- Quick Sort
- TO_DO Documentation
- File Exists
- Is Anagram
- Open File Debug
- Iterable has duplicates
- Iterable all elements unique


AndreaCrotti:
Sorry but some of these snippets are already there more or less (class) and others are just not really supposed to be snippets at all (all the sort algorithms).
If you use them often maybe better to keep them in your own snippets directory.

PR #485, opened 2024-01-27
Fix native-comp warnings in .yas-setup.el files

Hi-Angel:
There may be warnings in other files, I didn't look at those. I noticed two warnings in `.yas-setup.el` files, so decided to fix them 😊

The fact that they compile now without any new warnings was checked by a command:

```bash
emacs -batch --eval "(package-initialize) (setq byte-compile-error-on-warn t)" -f batch-byte-compile $(fnfx .yas-setup)
```
That revealed one doc-string warning, which I fix in a separate commit.


Hi-Angel:
Ping

Hi-Angel:
Ping

AndreaCrotti:
Nice thanks

PR #488, opened 2024-02-01
Fix yasnippet unknown directive warning

shassard:
While loading these snippets, the following warning is emitted: Ignoring unknown directive "cotributor" in file: ...


AndreaCrotti:
nice thanks

Issue #490, opened 2024-02-13
Two issues with `python-mode` snippets


joostkremers:
When I open a Python file, I get two yasnippet warnings. First:

    Ignoring unknown directive "NOTE" in file: /home/joost/.emacs.d/elpa/yasnippet-snippets-20240115.1138/snippets/python-mode/function_docstring

`function_docstring` has the following line in its header:

```
# NOTE: Use minimum indentation, because Emacs 25+ doesn't dedent docstrings.
```

Removing the colon (or the entire "NOTE: " part) solves the issue.

Second:

    Multiple snippets with same identity: "__iter__"

There are two snippets for `__iter__`, one in a file called `iter` and one in a file called `__iter__`. Unfortunately, both use `__iter__` for the `name:` directive. Since both snippets essentially do the same thing, perhaps one can be removed?

I know these are minor issues, but the warnings are displayed in the echo area and may push out other, more important, warnings. (Such as an issue with Eglot, in my case...)

swinkels:
I just had a look at the warning about the unknown directive "NOTE". That exact line has been in the snippet for 7 years so it must be something in yasnippet itself. Indeed, yasnippet commit https://github.com/joaotavora/yasnippet/commit/25f5d8808af23fb3b3dd6a7aacb06e17006ffca6 (19 January) added the warning message for unknown directives.

Removing the colon seems like a valid workaround, but of course a pull request to modify the snippet is better.

swinkels:
About the message for the multiple snippets with the same identity, I see that message is added in yasnippet commit https://github.com/joaotavora/yasnippet/commit/8e9ce76f7a6e1932f5a64f271eadfb2939a81128 (12 February). Looking at the code comment, it is allowed for one snippet to overwrite another one, but apparently this is not one of these cases.

swinkels:
@JulienPalard Your commit https://github.com/AndreaCrotti/yasnippet-snippets/commit/9f67370d9dbff73b0a4bf7c859455659b193f991 adds Python snippet file `__iter__`. This file adds a snippet with the name `__iter__`, just as the much older snippet file `iter` does. This leads to a "Multiple snippets with same identity" message. 

The snippets are slightly different. Is there a need to support both but if not, which should be the one to keep?

JulienPalard:
Thanks for reporting.

I opened pr #500 to remove one of the two iter. I kept the one consistent with the other snippets.

PR #499, opened 2024-05-17
Remove directive formatting from NOTE marker

swinkels:
Before this commit, this snippet formatted a NOTE marker as a snippet directive, as `NOTE:`. As `NOTE` isn't a directive, this triggered a directive warning.


swinkels:
This should address the first issue reported in #490.

PR #503, opened 2024-07-19
Added easycrypt-mode snippets

JoaoDiogoDuarte:
Easycrypt is a formal verification tool used in cryptography that takes inspiration from languages like Coq and OCaml.  

Since I use this language frequently, I added easycrypt snippets. They do not cover everything, but I believe they cover decent ground.


snippets/easycrypt-mode/#ecall#
```diff
@@ -0,0 +1,6 @@
+# -*- mode: snippet -*-
```
AndreaCrotti, line 1: why do you have files starting with `#` and why are there two files with the same name but one `#ecall#` and the other one is `#ecall` with just your name in it?

JoaoDiogoDuarte: Apologies, I must have forgotten to remove my emacs autosave files. I will go over these issues and tidy things up

PR #509, opened 2024-11-04
FIX: Python arguments.

### LINKED ISSUE #505: Any snippet using python-args-to-docstring returns Wrong type argument: listp
Maybe [issue #1200 from yasnippet](https://github.com/joaotavora/yasnippet/issues/1200) rather belongs here?  Possibly introduced by #465, 

> Expanding e.g. function_docstring and tabbing the cursor to args, shows error Wrong type argument: listp
> [...]

I can reproduce this with plain Emacs:

```py
def fd(foo):
    """
    Wrong type argument: listp, "foo"
    """

def fdn(bar):
    """

    Wrong type argument: listp, "bar"
    """ 
```

- GNU Emacs 29.4 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.38, cairo version 1.16.0) of 2024-07-22, modified by Debian
- yasnippet: 20241013.1557
- yasnippet-snippets: 20241014.949

quazgar:
- This should fix #505 "Any snippet using python-args-to-docstring returns Wrong type argument: listp".
- Add type hints to the docstring again.
- Remove `self` from arguments in docstring.

I also added a test in the setup file, but I don't suppose this is currently used in any pipeline?

### DIFF
--- a/snippets/python-mode/.yas-setup.el
+++ b/snippets/python-mode/.yas-setup.el
@@ -52,15 +52,20 @@ Groups:
   "Split python argument string ARG-STRING.
 
 The result is a list ((name, type, default), ...) of argument names, types and
-default values."
-  (mapcar (lambda (x)           ; organize output
-            (when (string-match python-split-arg-regex x)
-              (list
-               (match-string-no-properties 1 x) ; name
-               (match-string-no-properties 3 x) ; type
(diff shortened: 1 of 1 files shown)


AndreaCrotti:
Ah thanks for the tests, it's currently not set up but I can look into it

PR #525, opened 2026-03-12
Remove duplicate easycrypt-mode snippets

benthamite:
## Summary

- Three pairs of `easycrypt-mode` snippets share the same key, causing "Multiple snippets with same identity" warnings on `yas-reload-all`
- Removes the less complete duplicate from each pair, keeping the version with full metadata (`group`, `expand-env` headers) or standard naming

### Duplicates removed

| Kept | Removed | Shared key |
|------|---------|------------|
| `tactics/auto` | `tactics/auto.yasnippet` | `auto` |
| `tactics/do-split` | `tactics/dosplit` | `dosplit` |
| `misc/ge0` | `misc/geq0` | `ge0` |

## Test plan

- [x] Verified all three pairs have identical keys and equivalent template bodies
- [x] Confirmed kept files have equal or superior metadata (group, expand-env headers)
- [ ] CI passes (`lein test` in `report/`)


AndreaCrotti:
they are different snippets with different keys and different file names though

